PR #63, opened 2023-10-05
Make action dashboard columns customizable

bbliem:
Already in master, but still here is a PR for your reviewing enjoyment.


tituomin (approved):
Very good.

PR #65, opened 2023-10-07
Redact contact person information if plan is configured so

bbliem:
This takes care of the Asana task: https://app.asana.com/0/1202880927552397/1205365901296872

The contact information is at the moment still exposed by the `person` endpoint. Didn't we want to remove that altogether at some point?

### DIFF
--- a/people/models.py
+++ b/people/models.py
@@ -1,51 +1,48 @@
 from __future__ import annotations
 
 import contextlib
+import copy
 import hashlib
 import io
 import logging
 import os
 import re
-import uuid
-from datetime import timedelta
-from typing import TYPE_CHECKING, ClassVar
-
+import requests
 import reversion
+import uuid
+import willow
+from datetime import timedelta
 from django.contrib.auth.models import AnonymousUser
 from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import Q
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _, pgettext_lazy
+from easy_thumbnails.files import get_thumbnailer
+from image_cropping import ImageRatioField
 from modelcluster.models import ClusterableModel
 from modeltrans.fields import TranslationField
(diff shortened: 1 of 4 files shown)


people/models.py
```diff
@@ -384,6 +384,25 @@ def delete_and_deactivate_corresponding_user(self, acting_admin_user):
             target_user.deactivate(acting_admin_user)
         self.delete()
 
+    def get_redacted_copy(self, plan: Plan):
+        """Return a copy of self with redacted information according to the configuration of the given plan.
+
+        You better not save the returned object.
```
MWargelin, line 473: :sweat_smile: 

Would there be a way to enforce this?

I can think of this solution, is it too hacky?
```
def disable_saving():
    raise RuntimeError("Saving redacted copies of Persons not permitted.")

person = ...
person.save = disable_saving

return person
```

MWargelin (commented):
@bbliem Looks like I was added as a reviewer for this PR. Is it still relevant? Looks fine to me but Juha's comment maybe deserves some attention. Had one small thing to comment on.

people/models.py
bbliem: That would be one option. Or we could make a subclass of `Person`, or add a boolean flag `can_save` to `Person`, which achieves the same thing in a slightly less hacky way. But I don't think we need to be so concerned about the possibility of the user saving the returned object.

bbliem:
@juyrjola I addressed your comment about the dashboard action list by adding a field `hasContactPerson` to the responsible party node. The PR for the UI for this feature is here: https://github.com/kausaltech/kausal-watch-ui/pull/430

people/models.py
MWargelin: I guess it's not super likely we'd end up saving these, so I don't want to build huge defenses against that either. Creating the subclass sounds like the "correct" way to do this to me but also probably most laborious. I'm okay with anything, I'll close this thread and leave it up to you :stuck_out_tongue:

PR #74, opened 2023-10-13
Feature/category parent for reports

tituomin:
---
  - https://app.asana.com/0/0/1205383841109291


bbliem (commented):
I think it looks good overall, but I haven't tested it.

PR #85, opened 2023-11-03
Improve logging and uWSGI startup

juyrjola:
- Start using loguru for logging add a bit of logging context
- Perform Django system checks when starting under uWSGI


tituomin (approved):
There seems to be one extra import, otherwise ok!

PR #88, opened 2023-11-08
Add existing contact persons to "available for plan" person query

tituomin:
@bbliem FYI this fixes a bug causing the grid editor validation to fail because a certain plan contains test contact persons which are not actually "available for plan"


bbliem (approved):
Okay, why not...

PR #89, opened 2023-11-08
Admin: add a filter for showing people with same assignments as user

tituomin:


bbliem (approved):
Nice. I added the German translation.

PR #92, opened 2023-11-09
Disallow editors creating persons

tituomin:
People with only editor-level contact person rights do not have
the ability to add persons.


bbliem (approved):
Looks good. Is the `ActivatePermissionHelperPlanContextMixin` necessary?

PR #115, opened 2024-01-24
Add full width field to adaptive embed block

woodwoerk:
---
  - https://app.asana.com/0/0/1206390858072107


tituomin (approved):
Great!

PR #129, opened 2024-02-19
Implement various features for authenticated users in public UI

tituomin:


bbliem (commented):
I made some comments. Don't know if they make much sense because I'm kind of tired.

PR #133, opened 2024-02-27
Fix bug with target ancestor depth

bbliem:
I believe we had a bug in the code that displays the "containing organization" for a responsible party in a report based on the level in the hierarchy.

For example, with Bremen, who has a hierarchy with currently two levels, if you choose "1" as the target ancestor depth, then for a row whose responsible party is below the top level, you won't get its containing top-level organization but the organization itself.

I removed one of the conditional branches, namely the one handling `depth == 1` because I didn't see its use. It seemed to be based on the assumption that `ancestors` contains the organization itself, but it doesn't. Maybe I didn't see something.


tituomin (approved):
Good catch

PR #144, opened 2024-03-12
Added role field to the ActionContactPersoneNode

lilia1891:
Added a role field to be able to use it in GetActionList query and display it for a contact person in the admin ActionList table.

Related Asana task: https://app.asana.com/0/1203001306970200/1206371526193333/f

### DIFF
--- a/actions/models/action.py
+++ b/actions/models/action.py
@@ -932,7 +932,7 @@ class ActionContactPerson(OrderedModel, ModelWithRole):
     )
 
     public_fields: ClassVar = [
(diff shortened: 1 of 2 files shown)


tituomin (changes requested):
Thanks, I think it's better to add the field directly here:
https://github.com/kausaltech/kausal-watch-private/blob/master/actions/models/action.py#L935

tituomin (approved):
I suggest you "squash and merge", to reduce extra commits in master.

PR #145, opened 2024-03-12
Return draft attributes in GraphQL API

bbliem:
This PR is against `features/attributes-refactor`, which is awaiting separate review.


tituomin (approved):
Great!

PR #149, opened 2024-03-18
Fix bug: self.panels was mutated in SiteGeneralContentAdmin

tituomin:
Resulted in an extra field being added per language per field per new request


tituomin:
I have checked other usages of insert_model_translation_panels and in those contexts the panels gets added to tabs in the end, not to self.panels directly, so this bug doesn't appear there. Ie. the return value received from insert_model_translation_panels is not persisted but stored in local variables in all other cases. Also, insert_model_translation_panels creates a new list so even if in some cases we are sending self.*_panels as a parameter, that should be fine.

bbliem (approved):

PR #150, opened 2024-03-19
Fix hardcoded fi as default language when resolving gql i18n fields

tituomin:


aplans/utils.py
```diff
@@ -472,9 +474,30 @@ def get_supported_languages():
...
     return settings.LANGUAGES[0][0]
 
 
+def get_language_from_default_language_field(
+        instance: models.Model,
+        i18n_field: TranslationField | None = None
+):
+    """Return the primary language from the default language field"""
+
+    i18n_field = i18n_field or get_i18n_field(type(instance))
```
bbliem: Maybe better to use `instance._meta.model` instead of `type(instance)` to be sure the type of the argument is a model. But it should work like this as well.

bbliem (approved):

PR #154, opened 2024-03-22
Add action.visibility to report

Bhavatu:
Add action visibility as a field to reports

https://app.asana.com/0/1206017643443542/1206773871156898

### DIFF
--- a/reports/spreadsheets.py
+++ b/reports/spreadsheets.py
@@ -381,6 +381,10 @@ class ExcelReport:
             append_to_key(COMPLETED_BY_LABEL, completed_by or '')
             append_to_key(COMPLETED_AT_LABEL, completed_at)
             self.formats.set_for_label(COMPLETED_AT_LABEL, self.formats.timestamp)
+
+            if (visibility_label := action_obj.get_visibility_display()):
+                append_to_key(_('Visibility'), visibility_label)
+
         if data and set(data.get(COMPLETED_AT_LABEL)) == {None}:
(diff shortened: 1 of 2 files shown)


reports/spreadsheets.py
```diff
@@ -381,6 +381,12 @@ def append_to_key(key, value):
             append_to_key(COMPLETED_BY_LABEL, completed_by or '')
             append_to_key(COMPLETED_AT_LABEL, completed_at)
             self.formats.set_for_label(COMPLETED_AT_LABEL, self.formats.timestamp)
+
+            visibility_label = action.data.get('visibility', '')
+            if visibility_label:
+                visibility_label = str(dict(Action.VisibilityState.choices).get(visibility_label, visibility_label))
```
tituomin: You can use https://docs.djangoproject.com/en/dev/ref/models/instances/#django.db.models.Model.get_FOO_display

tituomin (dismissed):
This is a good minimal implementation, we should just be aware that this was requested by a specific customer and if we assume there will be some customers who might not want this (for example who don't use the visibility feature) then this will be an extra hardcoded column for them. But I have another feature pending which is a general solution where users can include/exclude built-in fields in reports, so we can merge this in the meanwhile.

tituomin:
You also need to change the test for the extra column (just add one)

reports/spreadsheets.py
Bhavatu: neat!

tituomin (approved):

PR #158, opened 2024-03-27
Indicators: internal_notes and reference field

Bhavatu:
Add internal notes field + public reference field for indicators

https://app.asana.com/0/1206017643443542/1206649411944555


Bhavatu:
As discussed with @terotik, Added the reference as a RichTextField instead to support embedding links, limiting features to links only and max_length=255

tituomin (approved):

PR #166, opened 2024-04-09
Handle missing attribute types in draft attributes

tituomin:


bbliem (approved):
It's probably not a problem at this point, but since there is a bit going on in `commit_attribute()`, it may be possible that this method raises `KeyError` as well, which we might not want to ignore. In general I like this pattern:
```
            try:
                attribute_value = draft_attributes.get_value_for_attribute_type(attribute_type)
            except KeyError:
                pass
            else:
                attribute_type.commit_attribute(self, attribute_value)
```

PR #168, opened 2024-04-10
Feature/report builtin field formatters

tituomin:


reports/spreadsheets/excel_report.py
```diff
@@ -454,7 +454,7 @@ def post_process(self, action_df: polars.DataFrame):
             aggregated = self._get_aggregates(grouping, action_df)
             if aggregated is None:
                 continue
-            sheet_name = f"Summary {sheet_number}"
+            sheet_name = _("Summary") + f" {sheet_number}"
```
bbliem, line 457: I'd prefer this to be something like `_("Summary %d") % sheet_number`

bbliem (commented):
Sorry, I had only a cursory look since the changes are huge and I think in order to give a useful review that really catches potential bugs, I'd have to spend an entire day or so. If you think that's necessary, I can revisit this.

PR #171, opened 2024-04-12
Bring back moderators to GQL API when authenticated to admin UI

Bhavatu:
Related Asana: https://app.asana.com/0/1206017643443542/1206886358160720

Depends on this: https://github.com/kausaltech/kausal-extensions/pull/19


tituomin (dismissed):
There's a problem which I didn't communicate properly yet:
In Watch nowadays, plan admins and other admin users can be logged into the public site. But in that case they are still counted as admin people, so the real distinction between frontend/backend use should be with another mechanism, maybe just a GraphQL parameter.

tituomin:
If I was not clear I meant:
`not user.can_access_admin(plan` < this one makes it so that those who can access admin will see the contact persons both front and back, but the intention was even those would only see it in the backend.

Bhavatu:
@tituomin note also the extensions PR in the description

tituomin (approved):

PR #176, opened 2024-04-18
Remove extra SQL queries of content types, related models with default languages, and action dependencies for plans not using the feature

tituomin:
Reduces the amount of queries from >9000 to ~599~  79 for that _one_ customer. All of the perf fixes for the action list page query are not done yet, but we should merge this ASAP and then work on the action dependencies.

### DIFF
--- a/actions/models/attributes.py
+++ b/actions/models/attributes.py
@@ -28,7 +28,6 @@ if typing.TYPE_CHECKING:
     from .plan import Plan
     from .category import CategoryType
     from actions.attributes import AttributeType as AttributeTypeWrapper, DraftAttributes
-    from actions.models import Action, Category
 
 
 class AttributeTypeQuerySet(models.QuerySet['AttributeType']):
@@ -190,13 +189,11 @@ class Attribute(models.Model):
     object_id = models.PositiveIntegerField()
     content_object = GenericForeignKey()
 
-    def is_visible_for_user(self, user: UserOrAnon, plan: Plan, model_with_attribute: Action | Category) -> bool:
+    def is_visible_for_user(self, user: UserOrAnon, plan: Plan) -> bool:
         from actions.models.action import Action
         assert plan is not None
         if self.content_type_id == ContentType.objects.get_for_model(Action).id:
-            assert self.object_id == model_with_attribute.pk
-            action = model_with_attribute
-            assert type(action) == Action
+            action = self.content_object
         else:
             action = None
         return self.type.is_instance_visible_for(user, plan, action)
--- a/actions/models/attributes.py
+++ b/actions/models/attributes.py
@@ -28,6 +28,7 @@ if typing.TYPE_CHECKING:
     from .plan import Plan
     from .category import CategoryType
     from actions.attributes import AttributeType as AttributeTypeWrapper, DraftAttributes
+    from actions.models import Action, Category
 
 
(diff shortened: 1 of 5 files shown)


actions/models/attributes.py
```diff
@@ -189,11 +190,13 @@ class Meta:
     object_id = models.PositiveIntegerField()
     content_object = GenericForeignKey()
 
-    def is_visible_for_user(self, user: UserOrAnon, plan: Plan) -> bool:
+    def is_visible_for_user(self, user: UserOrAnon, plan: Plan, model_with_attribute: Action | Category) -> bool:
```
bbliem: <strike>Well actually `model_with_attribute`  is an instance, not a model as the name would imply. Also, we have a base class called `ModelWithAttributes` for both `Action` and `Category`. How about something like`object: ModelWithAttributes` (or `instance`)  instead?</strike>

See comment below.

bbliem: Never mind my comment above. On second thought, I'm not sure I like this.

What's called `model_with_attribute` here is guaranteed to be the same instance as `self.content_object`. The reason that we supply it here as an argument seems to be solely that this way, we avoid hitting the DB in `action = self.content_object`.

I don't like that we have this redundant parameter. Sure, there is the assertion to enforce that `model_with_attribute` is in fact always `self.content_object`. But it's confusing to someone who looks at the method signature.

Luckily, we can solve the issue in a different way that does not require us to add a redundant parameter. We can add `<relation_name>__content_object` to the `prefetch_related` instruction in the `gql_optimizer.resolver_hints()` annotation of `resolve_attributes()`.

I tested this and it cuts down the number of queries similarly. Will add a commit to this PR...

(By the way, in this case it doesn't seem to apply here, but as of Django 5.0, there is [GenericPrefetch](https://docs.djangoproject.com/en/5.0/ref/contrib/contenttypes/#genericprefetch) that might come in handy in similar situations. We will be using Django 5.0 after the Wagtail upgrade PR is merged.)

actions/schema.py
```diff
@@ -635,7 +635,8 @@ class AttributesMixin:
     @gql_optimizer.resolver_hints(
         prefetch_related=[
             *chain(*[('%s' % rel, '%s__type' % rel) for rel in ModelWithAttributes.ATTRIBUTE_RELATIONS]),
-            *['choice_attributes__choice', 'choice_with_text_attributes__choice']
+            *['choice_attributes__choice', 'choice_with_text_attributes__choice'],
```
bbliem: If we prefetch `a__b`, then prefetching `a` seems to be redundant. That is, we can remove this line due to the next.

aplans/cache.py
```diff
@@ -16,6 +16,10 @@ def __init__(self, plan: 'Plan'):
     def action_statuses(self) -> list[ActionStatus]:
         return list(self.plan.action_statuses.all())
 
+    @cached_property
+    def plan_has_action_dependency_roles(self):
+        return self.plan.action_dependency_roles.count() > 0
```
bbliem: I prefer `.exists()` instead of `.count() > 0`.

bbliem (commented):
Good work. I made what I think is some improvements. Will add a commit of my own.

bbliem:
I would request a review from you now, but apparently this is not possible since it's your PR. So you can look at my commit and if you like it go ahead and merge.

tituomin:
Reviewed, approved, merged

PR #177, opened 2024-04-22
Add missing i18n panels to InlinePanel for CategoryLevel instances

bbliem:
As you can see on https://sunnydale.test.kausal.tech/de/climate/actions/T1, for example, the level "Theme" for the category type of the same name is displayed in English. We do support i18n for the `CategoryLevel` model, but there is no way to edit translations. This PR fixes that. It also makes the field "name_plural" translatable.


tituomin (approved):

bbliem:
I rebased this PR on the current `main` because there were some conflicts. Now there is some error in CI, possibly unrelated to this PR. I'll check it out after the holidays. Or if somebody wants to, be my guest.

PR #181, opened 2024-04-29
Create WIP budgeting data model

bbliem:
We also need editing facilities in Wagtail. And especially the REST API is a rather big task. (We can't just copy what we have in KP here because in KP things work quite differently with the dataframes and all.)

This is just for discussion / review, so probably better not merge this yet.

I have some code in my stash for the editing and API tasks by the way.

Asana task: https://app.asana.com/0/1206017643443542/1206938437282296/f


bbliem:
I took this up again and now we have a preliminary REST API and a modeladmin class for `budget.Dimension`.

tituomin (approved):

PR #182, opened 2024-04-29
Improve Accessibility: Add aria-expanded and aria-controls to action menu dropdown

Bhavatu:
`aria-expanded` and `aria-controls` added to action menu dropdown.
If i understood correctly, this was one of the things wished for.

Asana: https://app.asana.com/0/1206017643443542/1207179169863127

How it was tested:
Only manually by checking that the tags work as they should in Actions and Pages edit view. 
`aria-expanded` reflects if the expandable element the togglebutton controls is expanded or not, and `aria-controls` refers to the element that is controlled by the togglebutton.


Bhavatu:
Not needed after upgrade to wt6

PR #184, opened 2024-05-03
Delete more data with destructively_trim_db command

bbliem:
I had to create a migration that deletes old tables from django-orghierarchy. We removed this dependency a while ago but the tables still exist and they cause errors when we try to delete data that is referenced in django-orghierarchy tables.


MWargelin:
@bbliem This PR has been stale for a while, is it still relevant?

PR #185, opened 2024-05-06
Expose choice ids for choice type attributes in GQL

tituomin:
This is a performance improvement. The action list page is currently querying for all of the attributes and some unneeded fields for them, just to support filtering based on choice attributes. This implementation is easier to optimize.


tituomin:
@bbliem This implementation only works for published actions. I made this as a preliminary step because otherwise implementing these filters for my next PR, which uses draft action contents in the action list page, becomes more painful.

tituomin:
This of course implies a UI change which I will make a PR of if you find this ok.

tituomin:
Abandoning this approach for another one.

PR #186, opened 2024-05-06
Accessibility. Fix for screenreader to read out unsaved edits notification.

Bhavatu:
Change `.footer__container .footer__container--hidden`:  `display:none` --> `display:block`

With `none` the screenreaders ignore the changes that `aria-live` was set to detect

Depends on WT6 upgrade: https://github.com/kausaltech/kausal-watch-private/tree/chore/dependencies-upgrade

Tested manually with orca and NVDA.


bbliem (approved):
Thanks. I don't remember anymore why I set it to `none`. Deleting this rule doesn't change anything on the page I tested. Anyway, if you tested it with `block`, I trust it works.

PR #187, opened 2024-05-08
Do not show internal visibility actions in the public UI by default

tituomin:
Not even for authenticated users. Only when authenticated and explicitly requested (when calling from the admin UI)


tituomin:
Will need to include https://github.com/kausaltech/kausal-extensions/pull/22 after it has been released.

tituomin:
Reviewed and accepted separately.

PR #188, opened 2024-05-10
Implement efficient mass query of action revisions

tituomin:
Needed for action list page. Contains lots of optimizations.

### DIFF
--- a/actions/schema.py
+++ b/actions/schema.py
@@ -1167,11 +1167,11 @@ class ActionNode(AdminButtonsMixin, AttributesMixin, DjangoNode):
         only=('latest_revision', 'has_unpublished_changes', 'plan', 'plan__features__moderation_workflow'),
         prefetch_related=prefetch_workflow_states
     )
-    def resolve_workflow_status(root: Action, info) -> WorkflowState:
+    def resolve_workflow_status(root: Action, info) -> WorkflowState | None:
         user = info.context.user
         plan = root.plan
-        if not user.is_authenticated or not user.can_access_admin(plan):
-            raise ErrorCode.ACCESS_DENIED.create_error('Access denied')
+        if not user.is_authenticated or not user.can_access_public_site(plan):
+            return None
         return root
 
 
@@ -1354,11 +1354,13 @@ class Query:
     )
 
     workflow_states = graphene.List(
-        WorkflowStateDescription, plan=graphene.ID(required=True)
+        WorkflowStateDescription, plan=graphene.ID(required=False)
     )
 
     @staticmethod
     def resolve_workflow_states(root, info, plan):
+        if plan is None:
+            return []
         user = info.context.user
         result = []
         plan = Plan.objects.get(identifier=plan)
--- a/actions/schema.py
+++ b/actions/schema.py
@@ -1306,13 +1306,14 @@ def _resolve_action_revision(action: Action, desired_workflow_state: WorkflowSta
         revision_action._actual_workflow_state = match
         return revision_action
 
-    def published():
-        return with_workflow_state(WorkflowStateEnum.PUBLISHED, action)
+    def published(action: Action):
+        action._actual_workflow_state = WorkflowStateEnum.PUBLISHED
+        return action
 
     current_progress, max_progress = action.get_workflow_progress()
 
     if current_progress == max_progress:
-        return published()
+        return published(action)
 
     available_revision_state = WorkflowStateEnum.DRAFT
     if current_progress > 1:
@@ -1325,7 +1326,7 @@ def _resolve_action_revision(action: Action, desired_workflow_state: WorkflowSta
             return with_workflow_state(available_revision_state, action)
 
     # User wants published version or no other appropriate version available
-    return published()
+    return published(action)
 
 
 class Query:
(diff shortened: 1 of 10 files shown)


aplans/cache.py
```diff
@@ -37,6 +71,42 @@ def get_action_status(self, *, id: int | None = None, identifier: str | None = N
...
+            else:
+                if implementation_phase.identifier == identifier:
+                    return implementation_phase
+        return None
+
+    @cached_property
+    def attribute_choice_options(self) -> dict[int, AttributeTypeChoiceOption]:
+        result = {}
+        plan_content_type = ContentType.objects.get_for_model(Plan)
+        choice_formats = (
```
bbliem, line 89: If we were to add a new attribute format, we will likely forget that we need to add it here as well. Maybe better to identify the formats having choice options in the `AttributeType` class.

aplans/cache.py
```diff
@@ -37,6 +71,42 @@ def get_action_status(self, *, id: int | None = None, identifier: str | None = N
...
+            AttributeType.AttributeFormat.UNORDERED_CHOICE
+        )
+        for attribute_type in (
+            AttributeType.objects.filter(
+                scope_content_type=plan_content_type
+            ).filter(
+                scope_id=self.plan.pk
+            ).filter(
+                format__in=choice_formats
+            )
```
bbliem, line 101: ... or just:
```
AttributeType.objects.filter(
    scope_content_type=plan_content_type,
    scope_id=self.plan.pk,
    format__in=choice_formats,
)
```

actions/schema.py
```diff
@@ -1164,12 +1174,26 @@ class Meta:
...
 
 
 class ActionContactPersonNode(DjangoNode):
+    @staticmethod
+    def resolve_person(root: ActionContactPerson, info) -> Person:
+        cache = info.context.watch_cache.for_plan_id(root.action.plan_id)
+        person = cache.get_person(root.person_id) or root.person
+        person_organization = cache.get_organization(person.organization_id)
+        if person_organization is not None:
+            person.organization = person_organization
```
bbliem, line 1218: Here (and I believe in other places such as `enrich_action()`) I suppose it could happen that we override something that already contains an instance of a related object. Perhaps it would be a good idea to make sure that in this case at least the PK of what we override it with is the same as the PK of the thing that was there previously? For example, before we override `person.organization`, perhaps we should make sure that `person.organization.pk == person_organization.pk`.

actions/schema.py
```diff
@@ -1355,14 +1379,84 @@ def resolve_my_plans(root, info: GQLInfo):
...
+
     @staticmethod
     def resolve_plan_actions(root, info, plan, first=None, category=None, order_by=None, restrict_to_publicly_visible=True, **kwargs):
         plan_obj = get_plan_from_context(info, plan)
         if plan_obj is None:
             return None
-        qs = plans_actions_queryset([plan_obj], category, first, order_by, info.context.user, restrict_to_publicly_visible=restrict_to_publicly_visible)
-        result = gql_optimizer.query(qs, info)
-        return result
+        workflow_state = info.context.watch_cache.query_workflow_state
```
bbliem, line 1469: Maybe better put that down where it is first used.

actions/schema.py
```diff
@@ -1355,14 +1379,84 @@ def resolve_my_plans(root, info: GQLInfo):
...
+                first,
+                order_by,
+                info.context.user,
+                restrict_to_publicly_visible=restrict_to_publicly_visible
+            ),
+            info
+        )
+        user = info.context.user
+        cache = info.context.watch_cache.for_plan(plan_obj)
+        persons_queryset = Person.objects.filter(actioncontactperson__action__plan=plan_obj)
```
bbliem, line 1483: Could this be a problem because it hits the DB (in `populate_persons()`) every time this method is called, i.e., every time someone visits, say, the action list page?

actions/schema.py
```diff
@@ -1355,14 +1379,84 @@ def resolve_my_plans(root, info: GQLInfo):
...
+        cache.populate_organizations(
+            Organization.objects.filter(
+                Q(responsible_actions__action__plan=plan_obj) |
+                Q(people__in=persons_queryset)
+            )
+        )
+        if not is_authenticated(user):
+            workflow_state = WorkflowStateEnum.PUBLISHED
+        elif not user.can_access_public_site(plan=plan_obj):
+            workflow_state = WorkflowStateEnum.PUBLISHED
```
bbliem, line 1494: Since the logic is the same, how about:
```
        if not is_authenticated(user) or not user.can_access_public_site(plan=plan_obj):
            workflow_state = WorkflowStateEnum.PUBLISHED
```

actions/schema.py
```diff
@@ -1355,14 +1379,84 @@ def resolve_my_plans(root, info: GQLInfo):
...
+        actions_without_revision = []
+        if workflow_state == WorkflowStateEnum.DRAFT:
+            revision_pks = [pk for pk in action_queryset.values_list('latest_revision_id', flat=True) if pk is not None]
+            actions_without_revision = action_queryset.filter(latest_revision__isnull=True)
+        elif workflow_state == WorkflowStateEnum.APPROVED:
+            desired_workflow_task = plan.get_next_workflow_task(workflow_state)
+            action_pks = [str(pk) for pk in action_queryset.values_list('pk', flat=True)]
+            if desired_workflow_task is not None:
+                workflowstates = (
+                    WorkflowState.objects.active()\
```
bbliem, line 1439: The backslashes here should be redundant since the whole expression is in parentheses.

actions/schema.py
```diff
@@ -1355,14 +1379,84 @@ def resolve_my_plans(root, info: GQLInfo):
...
+                    .filter(content_type=ct)\
+                    .filter(current_task_state__task=desired_workflow_task)\
+                    .filter(object_id__in=action_pks)
```
bbliem, line 1442: One could use a single `filter()` call (see also below).

aplans/cache.py
```diff
@@ -71,3 +148,56 @@ def for_plan_id(self, plan_id: int) -> PlanSpecificCache:
...
+
+class OrganizationActionCountCache:
+    plans: list[Plan]
+    data: dict[int, int]
+    action_qs: ActionQuerySet
+    organization_responsible_party_queryset_filter: Q
+
+    def __init__(self, action_qs: ActionQuerySet) -> None:
+        self.action_qs = action_qs
+        self.organization_responsible_party_queryset_filter = Q()
```
bbliem, line 161: Why initialize this here when it will be overridden in `_construct_cache_data()`?

aplans/cache.py
```diff
@@ -71,3 +148,56 @@ def for_plan_id(self, plan_id: int) -> PlanSpecificCache:
...
+                    continue
+                result[org_id] = result.get(org_id, 0) + 1
+                organization_pks_from_revisions.add(org_id)
+        for action in actions_without_revisions:
+            for arp in action.responsible_parties.all():
+                result[arp.organization_id] = result.get(arp.organization_id, 0) + 1
+        self.organization_responsible_party_queryset_filter = Q(
+            Q(responsible_actions__action__in=self.action_qs) |
+            Q(id__in=organization_pks_from_revisions)
+        )
```
bbliem, line 187: I haven't checked it, but why would this be different than the following?
```
        self.organization_responsible_party_queryset_filter = (
            Q(responsible_actions__action__in=self.action_qs) |
            Q(id__in=organization_pks_from_revisions)
        )
```
That is, drop the outer Q?

bbliem (approved):
Thanks. I made some comments. If you think it works, go ahead.

PR #189, opened 2024-05-13
Add Indicator Causal Chain Block to action page

tituomin:
This needs a migration before we can merge the UI changes,
but I will add the block first to enable UI development.


Bhavatu (approved):
looks straightforward

PR #191, opened 2024-05-15
Visibility for indicators

Bhavatu:
Visibility setting for indicators for hiding indicators in public ui.

Asana: https://app.asana.com/0/1206692847418454/1206462481975808

Depends on this to reduce n+1:s : https://github.com/kausaltech/kausal-watch-ui/pull/374

Testing
=====
`indicators/test_schema.py`
and manually: Set indicator visibility to "internal". Check in the public UI that indicator is not visible in at least: 
Action page (action indicator, outcome map)
Indicator list
Indicator page (Indicator page for internal indicator not found, and not showing up in related indicator page cause or effect), Indicator insights graph
Search


actions/models/action.py
```diff
@@ -494,6 +477,13 @@ def get_previous_action(self, user: User) -> Action | None:
             .first()
         )
 
+    def get_visible_indicators(self, user: User):
+        return self.indicators.visible_for_user(user)
+
+    def get_visible_related_indicators(self, user: User):
```
tituomin: n+1 query? see my comment below

actions/schema.py
```diff
@@ -1044,6 +1046,14 @@ def resolve_next_action(root: Action, info):
     def resolve_previous_action(root: Action, info):
         return root.get_previous_action(info.context.user)
 
+    @staticmethod
+    def resolve_indicators(root: Action, info):
+        return root.get_visible_indicators(info.context.user)
+    
+    @staticmethod
+    def resolve_related_indicators(root: Action, info):
```
tituomin, line 1057: I don't remember, are we calling these in the UI for action lists, not just individual actions? I think there is an "indicators" column in the dashboard. So the question would be, should we see if this results in many queries per action? However, if it does, maybe it's not a new thing. Might be worth checking out though -- for example just load the action dashboard page with the old code and this new code, and compare the SQL query count in the logs.

tituomin (commented):
Looks good! There is a new thing to consider though which we probably didn't discuss when you started on this. Not a big deal, but:

Nowadays there is the "login to the public UI feature". Previously it was thought that publicly logged in users would see all the actions (and indicators) in the public watch UI that they see in the admin UI. However, it turned out that at least for one customer they would like to hide the "internal" actions completely from the public UI, since they are not considered a part of the whole action plan in sense, and users would find them confusing maybe.

This was the view of only one customer though, and it's not clear how it should work in general, but for now we added this special code for actions:
https://github.com/kausaltech/kausal-watch-private/blob/master/actions/schema.py#L311-L326
The reason we still want to sometimes show _all_ actions is that the same GraphQL API is also called from the admin UI -- hence it's based on a parameter. Later on we might want to have an explicit setting for this, to control if internal things are shown for public ui logged in users.

I'm kind of wondering, if we should for consistency have the same logic for indicators. Though I'm not sure if we are querying indicators from the admin UI currently.

Let's discuss this further and reach a decision.

Also see my comments about the DB query count.

indicators/schema.py
```diff
@@ -94,7 +94,10 @@ class IndicatorLevelNode(DjangoNode):
     class Meta:
         model = IndicatorLevel
         fields = public_fields(IndicatorLevel)
-
+    
+    @staticmethod
+    def get_queryset(self, info):
```
tituomin: maybe rename self to root for consistency and because it's a staticmethod

indicators/models.py
```diff
@@ -719,6 +744,18 @@ class Meta:
...
+
+    def visible_for_user(self, user: UserOrAnon | None) -> Self:
+        """ A None value is interpreted identically
+        to a non-authenticated user"""
+        if user is None or not user.is_authenticated:
+            return self.filter(indicator__visibility=RestrictedVisibilityModel.VisibilityState.PUBLIC)
+        return self
+    
+    def visible_for_public(self) -> Self:
+        return self.visible_for_user(None)
```
tituomin, line 757: This same code is in 3 places, it would probably be easy to extract into a base class, but we can live with it too for now.

tituomin (approved):

PR #194, opened 2024-05-27
Add excel macro to report excel to preprocess for printing

tituomin:


bbliem (approved):
Nice. But the actual macro is apparently compiled in the binary?

tituomin:
@bbliem Yes, there's no way but to author it in the Visual Basic editor in Excel and then extract it using a tool in xlsxwriter.

PR #195, opened 2024-05-29
Implement manually scheduled notifications

tituomin:
Manually scheduled notifications are one-off notifications to be sent
at the usual time for a plan on a specific date, and only once.

They have an embedded rich text content directly in the Template
and the notification engine generation logic is revised a little
bit to allow for this, and to work with the fact there can be 
multiple instances of these manually scheduled templates.

Also, the recipient generation differs somewhat from the current
automatic notifications. We are using the term "automatic
notifications" for notifications which are triggered automatically
when an action becomes stale for example, and "manually scheduled" (in
the code) aka. "single notifications" (in the UI) for these single-use
notifications.


notifications/models.py
```diff
@@ -330,20 +346,75 @@ def _get_contact_person_recipients(
...
+                    (Q(custom_email='') & Q(send_to_custom_email=False))
+                    | (~Q(custom_email='') & Q(send_to_custom_email=True))
+                ),
+                name='custom_email_iff_send_to_custom_email',
+            )
+        ]
+
+
+class ManuallyScheduledNotificationTemplate(NotificationTemplateBase):
+    base = ParentalKey(BaseTemplate, on_delete=models.CASCADE, related_name='date_based_templates', editable=False)
```
bbliem: Why `date_based_templates` and not, e.g., `manually_scheduled_templates`? Adding another term for this might be confusing. We already have two others ("manually scheduled" and "single").

notifications/models.py
```diff
@@ -330,20 +346,75 @@ def _get_contact_person_recipients(
...
+    date = models.DateField(null=False, blank=False)  # Must be interpreted as local to the plan timezone
+    content = RichTextField(verbose_name=_('content'), help_text=_('HTML content for the block'))
```
bbliem: The user might wonder what "the block" is. Also, they might not even be aware of what HTML is or how it comes into play here because all they see is a rich-text editor.

notifications/models.py
```diff
@@ -330,20 +346,75 @@ def _get_contact_person_recipients(
...
+                check=(
+                    (Q(custom_email='') & Q(send_to_custom_email=False))
+                    | (~Q(custom_email='') & Q(send_to_custom_email=True))
+                ),
+                name='custom_email_iff_send_to_custom_email',
+            )
+        ]
+
+
+class ManuallyScheduledNotificationTemplate(NotificationTemplateBase):
```
bbliem: It may be confusing that `ManuallyScheduledNotificationTemplate` is not a `NotificationTemplate`. Perhaps it would be clearer to rename `NotificationTemplateBase` to `NotificationTemplate` and rename what this PR calls `NotificationTemplate` to `PeriodicNotificationTemplate` or so.

notifications/models.py
```diff
@@ -330,20 +346,75 @@ def _get_contact_person_recipients(
...
+    content = RichTextField(verbose_name=_('content'), help_text=_('HTML content for the block'))
+
+    sent_notifications = GenericRelation('notifications.SentNotification', related_query_name='manually_scheduled_notifications')
+
+    send_to_action_contact_persons = models.BooleanField(verbose_name=_('send to action contact persons'), default=True)
+    send_to_indicator_contact_persons = models.BooleanField(verbose_name=_('send to indicator contact persons'), default=True)
+    send_to_organization_admins = models.BooleanField(verbose_name=_('send to organization admins'), default=True)
+
+    @property
+    def type(self):
```
bbliem: Apparently this is a hack so that we can use a `ManuallyScheduledNotificationTemplate` as if it were a `NotificationTemplate`, which (as of this PR) it is not. It may lead to problems that here `type` is not a field but a property. Why not just leave the `type` field in the base class? No harm if it always has the same value for all manually scheduled notifications. We might in the future then also add other manually scheduled notification types.

notifications/wagtail_admin.py
```diff
@@ -56,6 +58,28 @@ class BaseTemplateAdmin(AplansModelAdmin):
         FieldPanel('identifier')
     ]
 
+    def get_single_notification_panels(self, send_at_time):
```
bbliem: I suppose the word "single" is just used because we don't want to expose the user to the more technical term "manually scheduled". If so, it is perhaps better to keep the use of "single" to a minimum and *only* use that word in the translatable string, but not in the code.

PR #200, opened 2024-06-03
If client not found for one recipient send other notifications

tituomin:
Rationale: it is difficult to fully enforce that the client configuration is completely in sync with the person/user configuration all of the time. Hence, the probability of there sometimes being users without clients cannot be easily ruled out and that shouldn't prevent the sending of notifications to all the other recipients in a plan.


tituomin:
Fixes https://sentry.kausal.tech/organizations/kausal/issues/1567/events/5fa113c496d24f489d596d56af0d3d40/?project=3&query=is%3Aunresolved&referrer=previous-event&statsPeriod=24h&stream_index=0

bbliem (approved):

PR #203, opened 2024-06-04
Do not send notifications too early the first time they are sent

tituomin:
For a plan where notifications have not been sent before,
they were sent too early in the day because this function returned True
regardless of time of day.

---
  - https://app.asana.com/0/0/1207480512391393


Bhavatu (approved):
Looks right to me. If `daily_notifications_triggered_at` hasn't been set, function returns true only when `send_at_time` is reached.

PR #207, opened 2024-06-05
Feature/notification rescheduling

tituomin:
https://app.asana.com/0/1206017643443542/1207442965750714/f

The rationale for supporting this is that we need to handle changing of the manual notifications in some way.

The alternative would be to just lock the date field after it has been set, but that would have been bad if the user makes a mistake in the initial date selection. I guess we could have also allowed changing the date only for notifications where the date is still in the future compared to when the user tries to change the date (ie. only for not-sent notifications). Anyway, I suppose this flexibility introduced here (always support rescheduling) would be useful to users, for example for reusing the same notification text over and over without copypasting.

### DIFF
--- a/notifications/wagtail_admin.py
+++ b/notifications/wagtail_admin.py
@@ -1,6 +1,4 @@
-from typing import reveal_type
 from django.conf import settings
-from django.core.exceptions import ValidationError
 from django.forms import BaseFormSet, Select
 from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
--- a/notifications/wagtail_admin.py
+++ b/notifications/wagtail_admin.py
@@ -1,5 +1,7 @@
+from typing import reveal_type
 from django.conf import settings
-from django.forms import Select
+from django.core.exceptions import ValidationError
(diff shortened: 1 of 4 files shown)


notifications/wagtail_admin.py
```diff
@@ -15,13 +15,40 @@
...
+                    formset[i].add_error('date', _('Cannot schedule a notification for the past'))
+                continue
+            instance = item['id']
+            if new_date != instance.date:
+                # Rescheduling old notification
+                if new_date < local_current_date:
+                    formset[i].add_error('date', _('Cannot reschedule a notification for the past'))
+
+    def clean(self):
+        self._clean_manually_scheduled_notification_templates(self.formsets.get('manually_scheduled_notification_templates', None))
```
bbliem: Doesn't really matter, but why not like this?

```
    def clean(self):
        formset = self.formsets.get('manually_scheduled_notification_templates', None)
        if formset is not None:
            self._clean_manually_scheduled_notification_templates(formset)
```
And then you could get rid of the first two lines in the helper method (and `formset` would never be `None`).

bbliem (approved):

PR #208, opened 2024-06-07
User impersonation

Bhavatu:
User impersonation for superusers in admin

Allows an superuser to use the admin ui as another user.

https://app.asana.com/0/1206017643443542/1206820131128896

What was tested/ How to test
=====================
Tests: `users/tests/test_impersonation.py `

Manual: Go to People list in the dashboard as superuser. Click View as user for the user you want to impersonate.
When you want to quit, click "Back to own user" in the sidebar.

With non-superuser the button should not be visible and the endpoint forbidden, also impersonating another user when already impersonating is forbidden. Logging out from the site while impersonating releases the impersonation. 

Impersonator should be logged in `LoggedRequest` for every `['POST', 'PUT', 'PATCH', 'DELETE']` made while impersonating.


admin_site/permissions.py
```diff
@@ -25,3 +25,8 @@ def instances_user_has_any_permission_for(self, user, actions):
         collections = plan.root_collection.get_descendants(inclusive=True)
         qs = qs.filter(collection__in=collections)
         return qs
+
+
+def superusers_only_hijack(*, hijacker, hijacked):
```
juyrjola: Could you add type annotations here? E.g. `hijacker: User, hijacked: User`

admin_site/signals.py
```diff
@@ -0,0 +1,17 @@
...
+from django.dispatch import receiver
+from django.contrib import messages
+from hijack.signals import hijack_started, hijack_ended
+from django.utils.translation import gettext as _
+
+logger = logging.getLogger(__name__)
+
+@receiver(hijack_started)
+def on_hijack_started(sender, hijacker, hijacked, request, **kwargs):
+    logger.info(f"{hijacker} has started impersonation for user {hijacked}")
```
juyrjola: Might be good if these logs stood out a bit, so maybe initialize the logger like this:

```
from loguru import logger

hijack_log = logger.bind('impersonation)
```

Then you can supply the users also as structured data to make it easier to sift through the logs:

```
hijack_log.bind(impersonation_actor=hijacker.email, impersonation_target=hijacked.email).info(f"{hijacker} has started impersonation for user {hijacked}")
```

admin_site/signals.py
```diff
@@ -0,0 +1,17 @@
...
+    messages.warning(request, _(f"You are now viewing the site as {hijacked}"))
+
+@receiver(hijack_ended)
+def on_hijack_ended(sender, hijacker, hijacked, request, **kwargs):
+    logger.info(f"{hijacker} has ended impersonation for user {hijacked}")
```
juyrjola: Same here for logging. Also, type annotations should be added. The type for `request` should be `WatchAdminRequest`.

admin_site/templates/admin_site/release_impersonation.html
```diff
@@ -0,0 +1,13 @@
+<script>
+    $(document).ready(function() {{
+        $('.sidebar').append(
+            '<div class="release-impersonation">' +
+            '<form method="POST" action="{% url 'hijack:release' %}"">' +
+            '{% csrf_token %}' +
+            '<button type="submit" class="button bicolor icon icon-warning">Back to own User</button>' +
```
juyrjola: It might be good to use `{% trans ... %}` here already, although we don't need to provide actual translations until we roll the functionality out to plan admins.

admin_site/urls.py
```diff
@@ -1,11 +1,12 @@
-from django.urls import path
+from django.urls import include, path
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
 from .api import check_login_method
 
 urlpatterns = [
     path('login/check/', check_login_method, name='admin_check_login_method'),
+    path('hijack/', include('hijack.urls')),
```
juyrjola: I wonder if we could have the URL path as `impersonate/` without complications?

admin_site/signals.py
```diff
@@ -0,0 +1,17 @@
+import logging
+from django.dispatch import receiver
+from django.contrib import messages
+from hijack.signals import hijack_started, hijack_ended
+from django.utils.translation import gettext as _
+
+logger = logging.getLogger(__name__)
+
+@receiver(hijack_started)
+def on_hijack_started(sender, hijacker, hijacked, request, **kwargs):
```
juyrjola: Should have type annotations here

people/views.py
```diff
@@ -93,3 +93,36 @@ def post(self, request, *args, **kwargs):
             'days_until_expiration': int(settings.PASSWORD_RESET_TIMEOUT / (60 * 60 * 24)),
         })
         return redirect(self.index_url)
+
+class ImpersonateUserView(WMABaseView):
```
juyrjola, line 99: It might be good to start phasing out `ModelAdmin`-based views already. Maybe you could inherit from `wagtail.admin.views.generic.BaseOperationView` instead?

juyrjola: Having read through the rest of the code, I take this back – let's have it as it is. We'll convert this view, too, when we convert `PersonModelAdmin`.

Bhavatu: :+1:

bbliem (approved):

PR #209, opened 2024-06-07
Implement local media storage with S3 fallback

bbliem:
This is meant for all environments that are not production deployments.

### DIFF
--- a/aplans/settings.py
+++ b/aplans/settings.py
@@ -87,6 +87,7 @@ elif os.path.exists(os.path.join(BASE_DIR, '.env')):
     environ.Env.read_env(os.path.join(BASE_DIR, '.env'))
 
 DEBUG = env('DEBUG')
+DEPLOYMENT_TYPE = env('DEPLOYMENT_TYPE')
 ALLOWED_HOSTS = env('ALLOWED_HOSTS')
 INTERNAL_IPS = env.list('INTERNAL_IPS',
                         default=(['127.0.0.1'] if DEBUG else []))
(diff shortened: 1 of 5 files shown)


bbliem:
About this complaint...

> Argument 1 of "url" is incompatible with supertype "Storage"; supertype defines the argument type as "str | None"  [override]

I'm not sure the type specified in the stubs is correct. I don't see why it should be `None`.

juyrjola:
> About this complaint...
> 
> > Argument 1 of "url" is incompatible with supertype "Storage"; supertype defines the argument type as "str | None"  [override]
> 
> I'm not sure the type specified in the stubs is correct. I don't see why it should be `None`.

Yeah, the type annotations are not perfect. It very well can be that `None` is not a possible type. You can just ask mypy to ignore that line to get rid of the complain.

aplans/settings.py
```diff
@@ -14,11 +14,12 @@
 import sys
 import importlib.util
 from celery.schedules import crontab
-from typing import Literal
+from typing import Any, Literal
 
 import environ
 from corsheaders.defaults import default_headers as default_cors_headers  # noqa
 from django.utils.translation import gettext_lazy as _
+from environ.environ import ImproperlyConfigured, ParseResult, parse_qs
```
juyrjola: `environ` doesn't have type information, so you can add `"environ.*"` to the `tool.mypy.overrides` section in `pyproject.toml`.

juyrjola: `parse_qs` and `ParseResult` might be better imported from `urllib.parse`.

aplans/storage.py
```diff
@@ -0,0 +1,78 @@
+from django.core.files import File
+from django.core.files.storage import Storage
+from django.core.files.storage import FileSystemStorage
+from django.utils.deconstruct import deconstructible
+from storages.backends.s3boto3 import S3Boto3Storage
```
juyrjola, line 5: `storages.*` should probably also be added to mypy ignores.

aplans/storage.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, **settings):
+        self.s3_storage = MediaFilesS3Storage(**settings)
+        self.fs_storage = FileSystemStorage()
+
+    def _open(self, name: str, mode: str = 'rb') -> File:
+        # Since _save() always works on the disk, we don't want to open S3 files for writing. Then again, is _open()
+        # actually ever called for anything else but reading?
+        assert mode == 'rb'
+        try:
+            return self.fs_storage._open(name, mode)
```
juyrjola: The warnings about `_open()` and `_save()` seem to be more about deficiencies in the type stubs, so you can safely `# type: ignore` them.

PR #213, opened 2024-06-18
Fix notification type mismatch with new manual notifications

tituomin:
Fixes a bug preventing creating plans and adding default notification contents.

Note, there are still some type problems in the file notifications.py, I would like to make the base class Notification generic wrt. the obj attribute, but better get this out soon and do that later.

### DIFF
--- a/notifications/management/commands/initialize_notifications.py
+++ b/notifications/management/commands/initialize_notifications.py
@@ -37,64 +37,12 @@ def initialize_notification_templates(
         'font_family': None,
         'font_css_url': None
     }
-    default_intro_texts_for_notification_type = {
-        'action_not_updated': pgettext(
-            'action_not_updated',
-            "This is an automatic reminder about "
-            "updating the action details in the action plan.  You can "
-            "see on the action plan watch site what has already been done "
-            "to further the actions and what has been planned for the "
-            "future.  It's already six months since you last updated an "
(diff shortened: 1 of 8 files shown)


tituomin:
> This bugs prevented creating plans and adding default notification contents.
> 
> Note, there are still some type problems in the file notifications.py, I would like to make the base class Notification generic wrt. the obj attribute, but better get this out soon and do that later.

The mypy check failing is related to this, let me know if you think I should fix it within this PR. There are other old similar type errors related to the fact the obj type changes (for user feedback objects, at least)

notifications/management/commands/initialize_notifications.py
```diff
@@ -37,61 +37,20 @@ def initialize_notification_templates(
...
+            continue
+
+        defaults = {
+            'subject': notification_type.verbose_name,
+        }
+        template, created = AutomaticNotificationTemplate.objects.get_or_create(
+            base=base_template, type=notification_type.identifier, defaults=defaults)
+        ContentBlock.objects.get_or_create(
+            template=template, identifier='intro', base=base_template,
+            defaults={'content': split_into_draftail_paragraphs(default_intro_text)})
```
bbliem: Closing parentheses like that aren't really our code style, but I don't have strong feelings about this.

bbliem (approved):

PR #216, opened 2024-06-26
Tweak Spectacular schema generation to work around typeconv bug

tituomin:
A null is a valid value for an enum but breaks our client side
schema->typescript conversion.

Spectacular automatically adds NullEnum to the oneof of the date_format type
which is correct but crashes our typeconv in extensions

### DIFF
--- a/aplans/settings.py
+++ b/aplans/settings.py
@@ -501,6 +501,10 @@ SPECTACULAR_SETTINGS = {
     'SERVE_INCLUDE_SCHEMA': False,
     'SCHEMA_PATH_PREFIX': '^/v1',
     'SCHEMA_COERCE_PATH_PK_SUFFIX': True,
+    # Strictly speaking, this results in the wrong type for eg. Action.date_format,
+    # but js typeconv cannot handle the null value for an enum
+    # https://github.com/grantila/typeconv/issues/39
+    'ENUM_ADD_EXPLICIT_BLANK_NULL_CHOICE': False,
     'ENUM_NAME_OVERRIDES': {
         'OtherLanguagesEnum': LANGUAGES,
     },


MWargelin:
Is this still relevant? :) It would be nice to get this PR cleaned up from the kuumotus-bot list

tituomin:
I have decided not to merge this. But the problem still remains:
DRF-spectacular correctly emits `null` as one possible value for a nullable Enum. But there is a bug in typeconv which can't handle the `null` value though it _should_ be there according the OpenAPI specs: https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2019-10-31-Clarify-Nullable.md#if-a-schema-specifies-nullable-true-and-enum-1-2-3-does-that-schema-allow-null-values-see-1900

I decided having faulty specs would be a bigger problem, eg. for integrations, than not having the typescript type generation work.

So currently if you try generate the types based on the OpenAPI schema for the REST API (in extensions), you get this:

```$ npm run watch-codegen

> kausal-watch-extensions@0.1.0 watch-codegen
> graphql-codegen -c codegen-watch.ts && node convert-rest-schema.mjs src/watch/generated/rest-api.ts

✔ Parse Configuration
✔ Generate outputs
Fetching from http://localhost:8000/v1/schema/...
Converting to TypeScript...
file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:205
    const { properties, additionalProperties = false, } = node;
            ^

TypeError: Cannot destructure property 'properties' of 'node' as it is null.
    at tsObjectType (file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:205:13)
    at tsConstType (file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:185:31)
    at file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:139:30
    at Array.map (<anonymous>)
    at tsType (file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:139:18)
    at tsTypeAndOrSchema (file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:109:16)
    at file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:100:76
    at Array.map (<anonymous>)
    at tsTypeUnion (file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:100:48)
    at tsTypeAndOr (file://__________________extensions/client/node_modules/core-types-ts/dist/lib/core-types-to-ts.js:115:16)
```

This should really be fixed in typeconv and/or core-types-ts.

PR #224, opened 2024-07-05
Add ordering to dimension categories of dataset schemas

tituomin:


budget/api.py
```diff
@@ -37,12 +37,20 @@ def __init__(self, *args, **kwargs):
 
 class DimensionCategorySerializer(I18nFieldSerializerMixin, serializers.ModelSerializer):
     # Reference dimension by UUID instead of PK
-    # dimension = serializers.SlugRelatedField(slug_field='uuid', read_only=True)  # implicit as router is nested
+    dimension = serializers.SlugRelatedField(slug_field='uuid', read_only=True)  # implicit as router is nested
```
bbliem: Can you remove the comment? It was there to explain why the line was commented out -- because I saw no need for the dimension UUID because it was already visible in the URL when accessing a dimension category.

bbliem (approved):

PR #226, opened 2024-07-09
Migrate site general content to snippets

MWargelin:


content/wagtail_admin.py
```diff
@@ -1,89 +1,79 @@
...
-        # The overridden superclass method returns True iff user_can_list from the permission helper returns true. But
-        # this menu item is about editing a plan, not listing.
-        plan = request.user.get_active_admin_plan()
-        return self.model_admin.permission_helper.user_can_edit_obj(request.user, plan.general_content)
 
+class SiteGeneralContentEditView(SuccessUrlEditPageMixin, AplansEditView):
+    permission_policy: SiteGeneralContentPermissionPolicy
 
-class SiteGeneralContentEditView(SuccessUrlEditPageMixin, SetInstanceMixin, EditView):
-    pass
```
MWargelin: The `SiteGeneralContentEditView` used to use `SetInstanceMixin`, but I left it out because as far as I could tell there was always access to the instance when needed anyway. But is there some larger idea behind setting the instance to the context that I don't see?

content/wagtail_admin.py
```diff
@@ -1,89 +1,79 @@
...
 
 # FIXME: This is partly duplicated in actions/wagtail_admin.py.
-class SiteGeneralContentPermissionHelper(ActivePlanPermissionHelper):
-    def user_can_edit_obj(self, user, obj):
-        return user.is_general_admin_for_plan(obj.plan)
-
+class SiteGeneralContentPermissionPolicy(ModelPermissionPolicy):
+    def user_has_permission(self, user, action):
+        if action == 'view':
+            return user.is_superuser
```
bbliem, line 19: I haven't checked what exactly the "view" permission is for, but I suppose plan admins that are not superusers are not restricted by this in practice since they have "change" permission?

content/wagtail_admin.py
bbliem: I don't think so. If the edit view works, it should be fine.

admin_site/viewsets.py
```diff
@@ -0,0 +1,42 @@
...
+from django.db import models
+from wagtail.snippets.views.snippets import CreateView, EditView, SnippetViewSet
+
+from admin_site.forms import WatchAdminModelForm
+from aplans.types import WatchAdminRequest
+
+M = TypeVar('M', bound=models.Model)
+
+
+class AplansEditView(EditView, Generic[M]):
```
bbliem: As discussed on Slack, perhaps it's a good idea to rename the other `AplansEditView` in `admin_site.wagtail` to something that expresses that it's legacy code until the migration away from modeladmin is done. Same for `AplansCreateView` and `SuccessUrlEditPageMixin`.

bbliem: Actually, another idea, the name "aplans" is only used in some places for historic reasons. For example the class `WatchAdminRequest` already uses the new name "watch". Perhaps this is a good opportunity to replace "Aplans" by "Watch" in the new classes.

content/wagtail_admin.py
```diff
@@ -1,89 +1,79 @@
...
+        TranslatedFieldPanel('official_name_description'),
+        TranslatedFieldPanel('copyright_text'),
+        TranslatedFieldPanel('creative_commons_license'),
+        FieldPanel('github_api_repository'),
+        FieldPanel('github_ui_repository'),
+        FieldPanel('action_term'),
+        FieldPanel('action_task_term'),
+        FieldPanel('organization_term'),
+        FieldPanel('sitewide_announcement'),
+    ]
```
bbliem: It would be kind of nice if we didn't have to change `FieldPanel` to `TranslatedFieldPanel` whenever we add a field to `i18n` in the model. In the old code, we don't have to do this because `insert_model_translation_panels` takes care of adding panels for all translatable fields. I haven't thoroughly thought about a solution yet though.

Maybe we could use a single class to be used instead of both `FieldPanel` and `TranslatedFieldPanel` that renders different things based on whether the field is in `i18n` or not.

Or maybe we could specify this in a different attribute than `panels` and then implement some logic that checks which of them are in `i18n` and populates `panels` with `FieldPanel` or `TranslatedFieldPanel` accordingly. Or instead of populating `panels` maybe there is a method `get_panels()` that we can override.

admin_site/panels.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def clone_kwargs(self):
+        ret = super().clone_kwargs()
+        del ret['children']
+        ret['field_name'] = self.field_name
+        ret['widget'] = self.widget
+        return ret
+
+
+class TranslatedFieldPanel(MultiFieldPanel):
```
bbliem: The content of this class is the same as for `TranslatedFieldRowPanel`. Perhaps it would be a good idea to put the common code (i.e., everything) into a mixin and reduce both class bodies to `pass`.

bbliem (commented):
Looks pretty good. Please see my comments for some things that may be worth considering.

content/wagtail_admin.py
MWargelin: alright, good :+1:

content/wagtail_admin.py
MWargelin: I think the `'view'` permission is used to determine both can the user see the index view and the inspect view. The end result here is that plan admins that are not superusers ~~do not see the Site content menu icon at all and are not given access, which seems to be how it works in the old code as well. But was it wrong in the first place then?~~

Edit: I checked it completely wrong, so actually plan admins that are not superusers still can edit the Site settings, both in old and new code. Those who are not plan admins are not given any access.

admin_site/viewsets.py
MWargelin: Good to know, I've been a bit confused about "aplans" being used in some places. I now renamed the legacy classes adding `ModelAdmin` somewhere in the name, that should be easy to find when trying to find last remaining modeladmin references. I also replaced `Aplans` with `Watch` in the new classes.

admin_site/panels.py
MWargelin: I did that but that caused some issues with type checking since the mixin reaches for superclass and assumed that the superclass has certain attributes. I didn't know what to do with it so I simply ignored the errors with `# type: ignore`: https://github.com/kausaltech/kausal-watch-private/pull/226/commits/c55927209c7312a72c9f4da5de5b7a8f7b569855#diff-1c5980c59a61704b9eab1c728753486b59e1a02fbfe912d500c1661c83004b14R51. Do you have a better solution? If not, do you still think it's better this way or should I go back and not have the mixin?

content/wagtail_admin.py
MWargelin: I did that now in https://github.com/kausaltech/kausal-watch-private/pull/226/commits/e78fcabbb7b15a663614cc327e6f9a4310251947 by using the `@property` decorator to have a bit of programmatic logic to apply the classes to the i18n fields. Sadly it's a custom solution for this class, I'm not entirely sure what would be a sensible way to make it generic...

bbliem: I don't really know a better way right now either. But I guess we can leave it like this for now.

admin_site/panels.py
bbliem: I think it's fine this way. Mixins don't work well with Python type checking. But I'd rather have a little bit less of type checking than larger amounts of duplicate code.

PR #227, opened 2024-07-16
Migrate modeladmin classes to snippets

MWargelin:
Migrates notification settings and plan features to snippets.

As the original classes were using `AplansEditView`, the new `WatchEditView` was updated to mimic the original as closely as possible.

### DIFF
--- a/actions/wagtail_admin.py
+++ b/actions/wagtail_admin.py
@@ -11,7 +11,6 @@ from wagtail.admin.panels import (
 from wagtail.snippets.models import register_snippet
 from wagtail_modeladmin.helpers import PermissionHelper
 from wagtail_modeladmin.options import modeladmin_register, ModelAdminMenuItem
-from wagtail_modeladmin.views import EditView
 
 from aplans.types import WatchAdminRequest
 
@@ -23,8 +22,7 @@ from actions.chooser import CategoryTypeChooser, PlanChooser
 from actions.models.action import ActionSchedule
 from admin_site.wagtail import (
     ActivePlanEditView, AplansAdminModelForm, AplansModelAdmin,
-    CondensedInlinePanel, SuccessUrlEditPageModelAdminMixin,
-    insert_model_translation_panels
+    CondensedInlinePanel, insert_model_translation_panels
 )
 from aplans.context_vars import ctx_instance, ctx_request
 from notifications.models import NotificationSettings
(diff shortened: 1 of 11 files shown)


notifications/models.py
```diff
@@ -52,6 +52,7 @@ def notification_type_choice_builder(include_manual: bool = False):
         yield (val.identifier, val.verbose_name)
 
 
+@reversion.register()
```
MWargelin, line 55: Seemed like this had to be added for the editing to work after snippet migration. I don't know why this wasn't needed earlier, is adding this ok or should I investigate further?

actions/wagtail_admin.py
```diff
@@ -283,7 +283,10 @@ def get_queryset(self, request):
 
 
 # FIXME: This is partly duplicated in content/admin.py.
-class ActivePlanPermissionHelper(PermissionHelper):
+# TODO: Reimplemented in admin_site/permissions.py to make this work without
+# ModelAdmin. Use that when implementing new classes or migrating away from
+# ModelAdmin. Remove this class when ModelAdmin migration is finished.
+class ActivePlanModelAdminPermissionHelper(PermissionHelper):
```
MWargelin, line 289: The old code doesn't seem to let plan admins to change the notification settings or plan features, even though the permission helper here looks like the idea is to let plan admins edit as well. I think this is a bug with `user_can_edit_obj`, where `obj` is sometimes a `Plan` instance but in theses cases it's not, so this time what we want would be `user.is_general_admin_for_plan(obj.plan)` instead of `user.is_general_admin_for_plan(obj)`.

Anyways, in the new code I tried to fix this bug in `ActivePlanPermissionPolicy` and plan admins can edit the notification settings and plan features, is that how it should be?

admin_site/mixins.py
```diff
@@ -12,3 +25,101 @@ def get_success_buttons(self) -> list:
...
+        filter_qs = self.request.session.get(f'{model}_filter_querystring')
+        if filter_qs is None:
+            return url
+        # Notice that urljoin will just overwrite any existing query
+        # strings in the url.  The query strings would have to be
+        # parsed, merged, and serialized if url contains query strings
+        return urljoin(url, filter_qs)
+
+
+class ContinueEditingMixin:
```
MWargelin, line 68: I didn't investigate this very deeply yet but it seemed like "save and continue editing" might be modeladmin specific feature and these continue editing related mixins might not be relevant anymore. Do you @bbliem happen to know already if this is the case or should I look more into this?

actions/wagtail_admin.py
bbliem: I'm not sure what the intention was. But if in the past plan admins have not been able to edit notification settings and plan features, then I think we should keep it like this. They won't be expecting this to work if it never worked. It may cause confusion if something new with potentially quite advanced and possibly unclear possibilities pops up. Also, there might be bugs or things that we did not intend, and we don't want users to trigger that.

actions/wagtail_admin.py
```diff
@@ -348,88 +354,81 @@ def get_menu_item(self, order=None):
...
-    ]
-
     panels = [
         FieldPanel('enable_search'),
         FieldPanel('enable_indicator_comparison'),
         FieldPanel('indicator_ordering'),
+        # Arbitrary string as the 'permission' parameter, here 'superuser', can
+        # be used as a way to restrict a panel only to superusers. This is the
+        # recommended approach given in Wagtail docs as of writing:
+        # https://docs.wagtail.org/en/v6.1.3/reference/pages/panels.html#wagtail.admin.panels.FieldPanel.permission
```
bbliem, line 367: Cool. I'm glad they implemented this.

admin_site/permissions.py
```diff
@@ -25,3 +32,79 @@ def instances_user_has_any_permission_for(self, user, actions):
...
+        if action == 'delete':
+            return False
+        return super().user_has_permission(user, action)
+
+    def user_has_permission_for_instance(self, user, action, instance):
+        if action == 'change':
+            if isinstance(instance, Plan):
+                return user.is_general_admin_for_plan(instance)
+            else:
+                return user.is_general_admin_for_plan(instance.plan)
```
bbliem: I could imagine two issues here:
1. This is called `ActivePlanPermissionPolicy`, but in this method we do not care whether the plan is the *active* plan. Maybe we should. I'm not totally sure, but if it does not break anything, perhaps it might be a reasonable safeguard to only be able to change something if its plan is also active. Can you think of a reason why this would be a bad idea? If no, could you try what happens when we check whether the plan as in fact also active, in addition to checking whether the user is an admin for it?
2. Some of our models are connected to a plan by more complicated means than just having a `plan` attribute. For example, `CategoryType` works fine with this code because it has a `plan` attribute. But `Category` does not; it is linked to a plan via its `type` attribute (a `CategoryType`). For this reason, as you know, there is a model called `PlanRelatedModel` with the method `get_plans()`. This returns all plans an object is connected to. In these cases, this will be just one plan, but in some cases (`Indicator`) it could be more. `CategoryType`, for example, uses the default implementation of `get_plans()`, which just returns `[self.plan]`. `Category` overrides this, for reasons I described before, and returns `[self.type.plan]`. Therefore, perhaps it's more reasonable to change the "else" branch so that we check whether `instance` is an instance of `PlanRelatedModel`, and if so, get the plan from `get_plans()`. (I suppose the permission should be granted if the user is an admin for any of the returned plans -- perhaps additionally requiring that this plan is active as mentioned above.) If `instance` is neither a `Plan` nor a `PlanRelatedModel`, perhaps just return `False` or panic (`assert False` or something like this). Below, there is the class `PlanRelatedPermissionPolicy`. Doesn't this do (or is supposed to do) much of what `ActivePlanPermissionPolicy` does?

notifications/models.py
bbliem: It should be fine to add this.

bbliem (commented):
Thanks! Looks good I think, but I haven't tested it, and this should really be tested. I suppose you did? If things work and you have dealt with my comments in one way or another, feel free to merge it.

notifications/models.py
MWargelin: great :+1:

actions/wagtail_admin.py
MWargelin: I feel like the intention must have been for plan admins to be able to edit at least plan features, since the code is making extra effort to make some panels in the edit view available only for superusers: https://github.com/kausaltech/kausal-watch-private/blob/a7d9c8e25e57d8f0ae8b0ea9575e9f84091bd9a7/actions/wagtail_admin.py#L357

But I agree that it might be risky to change this here and customers aren't waiting for this anyway. This should probably be checked though, would someone like @juyrjola or @tituomin know?

MWargelin: I now dug into the commit history and found commit c259fa91d0879a918fb5fe088b39e892e4677b35 . Looks like plan features used to have a dedicated permission helper `ActivePlanFeaturesPermissionHelper` that was changed in that commit to `ActivePlanPermissionHelper`. The old `ActivePlanFeaturesPermissionHelper` had `user.is_general_admin_for_plan(obj.plan)` in the `user_can_edit_obj` method, but now by switching to `ActivePlanPermissionHelper` that would have effectively been changed to `user.is_general_admin_for_plan(obj)`, making the method not quite work with a `PlanFeatures` instance. Looks like maybe plan admins were denied access accidentally here?

All in all (given your other comment https://github.com/kausaltech/kausal-watch-private/pull/227#discussion_r1681598103 as well) the permission helpers are a bit confusing right now... I'm thinking if for the plan features and notification settings a more fitting permission helper / permission policy would be something like `PlanSpecificSingletonModelPermissionPolicy` and some custom rules associated with it, rather than `ActivePlanPermissionHelper`. Maybe `ActivePlanPermissionHelper` should be dedicated to `ActivePlanAdmin` alone?

admin_site/mixins.py
bbliem: I have no idea unfortunately.

actions/wagtail_admin.py
bbliem: I don't know. I think at this point you have a better overview than I do. As long as things appear to work, I trust your judgment.

MWargelin: I now made commit a19e37b6d5186b5abb815da52fcc4514e380d5b2 trying to convey what we discussed in daily. What do you think?

admin_site/permissions.py
MWargelin: If we settle for a19e37b6d5186b5abb815da52fcc4514e380d5b2 we actually don't need these class at all (yet), so this maybe becomes an issue to solve in the future. Some thoughts anyway at this point:

- Sounds sensible to check for the active plan. I don't really have the overview to say whether it ruins something but sounds good.
- Good to know anyway about the intricacies of finding out the plan, I didn't know about these.
- I agree that `PlanRelatedPermissionPolicy` and `ActivePlanPermissionPolicy` look a lot like each other. Looks like the code could (and should) be refactored to use only one of them, but I feel like it would be easiest to do that separate from the already messy modeladmin migration...

actions/wagtail_admin.py
bbliem: Looks good to me.

admin_site/permissions.py
bbliem: Okay, let's keep this in mind then.

admin_site/mixins.py
MWargelin: Let's merge this as is now and see later if this becomes relevant or not.

PR #230, opened 2024-08-02
Budget dataset block and chooser

Bhavatu:


actions/blocks/action_content.py
```diff
@@ -240,6 +241,38 @@ class Meta:
         label = _("Indicator Causal Chain")
 
 
+
+class BaseDatasetsBlock(blocks.StructBlock):
+    field_label = blocks.CharBlock(
+        required=False,
+        help_text=_("What label should be used in the public UI for the official name?"),
```
tituomin: This looks like a copy-paste leftover. (offical name?)

tituomin:
@Bhavatu You should probably run makemigrations and add the migrations, even though not critical in this case.

actions/blocks/action_content.py
Bhavatu: Yep. Fixed

budget/models.py
```diff
@@ -90,6 +90,7 @@ class TimeResolution(models.TextChoices):
     unit_i18n: str
     name_i18n: str
 
+
     def __str__(self):
         if self.name_i18n:
             return f'{self.name_i18n} ({self.uuid})'
```
tituomin, line 96: It looks like the chooser is using this as the schema name in the UI, I find the UUID a bit ugly but would perhaps like to keep this `__str__` for internal developer debugging purposes where the uuid is probably helpful. Not critical, but if it's easy to use just the name_i18n for the chooser, we might do it.
Screenshot from 2024-08-02 14-59-09

tituomin (approved):

tituomin (approved):

PR #231, opened 2024-08-05
Expose dataset values as numbers, not strings in budget REST API

tituomin:


bbliem (approved):

Bhavatu (approved):
Does what is described. I don't know about downstream implications of this, but should be safe.

PR #238, opened 2024-08-08
Feedback to support category as feedback type

Bhavatu:
Feedback to support category as feedback type. also some ruffing

Relevant UI PR: https://github.com/kausaltech/kausal-watch-ui/pull/390

Asana: https://app.asana.com/0/1206017643443542/1207572329789663


pages/models.py
```diff
@@ -127,9 +127,6 @@ class AplansPage(Page):
     class Meta:
         abstract = True
 
-    @property
-    def preview_modes(self):
-        return []
```
tituomin, line 130: I'm curious about this, but I trust you checked this doesn't break something (at least obviously)

tituomin (approved):

pages/models.py
Bhavatu: It is a duplicate of https://github.com/kausaltech/kausal-watch-private/pull/238/files/d28abc73f2c3a027583ddf89545da261ab9a737f#diff-6dc505612eddf1930328c8438e149fc01ee563d4116ed7ecbb289d5072f82ddaR161-R163 from the same class that ruff was complaining about. I haven't noticed anything breaking from it

PR #239, opened 2024-08-08
Only allow choosing modifiable actions as actions related to indicators

bbliem:
This adds the option `only_modifiable` to `ActionAutocomplete` for this purpose.


bbliem:
Luckily, this does not erase existing action-indicator links even if the editing user is not permitted to modify the linked action. The linked action will stay and be displayed to the user. It's just that they cannot select it in the chooser anymore.

tituomin (approved):

MWargelin:
@bbliem looks like this has been forgotten, is this still relevant?

PR #243, opened 2024-08-15
Fix saving for models which are not registered with revision

tituomin:
When recently fixing the missing revision saving logic for actions,
the execute_admin_post_save_tasks was taken into use also for models
which are not registered with reversion. This affected at least Client
models.


Bhavatu (approved):
Looks good. 
(Note ruff complaints about unsorted imports)

PR #249, opened 2024-08-27
Indicator improvements

MWargelin:
- Reorganize indicator edit view
- Add option to show/hide trend line in indicator visualization
- Add option to determine whether indicator visualizations should show something increasing or decreasing as desired improvement


indicators/wagtail_admin.py
```diff
@@ -382,123 +383,184 @@ class IndicatorAdmin(AplansModelAdmin):
...
-                    'contact_persons',
-                    panels=[
-                        FieldPanel('person', widget=PersonChooser),
-                    ],
-                ),
-            ], heading=_('Contact persons')),
+
+        # Advanced settings
+        advanced_panels: list[Panel] = [
+            FieldPanel('organization', widget=autocomplete.ModelSelect2(url='organization-autocomplete')),
```
MWargelin, line 459: organization and common indicator were forgotten from the from the plan how to organize the edit view. I now added organization to the advanced settings panel and common indicator to the relationships tab, do those places make sense?

MWargelin:
Here are some screenshots @terotik :

Basic information tab (some sections are collapsed by default):
Screenshot 2024-09-09 at 13-21-32 Editing indicator - Climate - Administration

Collapsed sections opened up:
Screenshot 2024-09-09 at 13-21-48 Editing indicator - Climate - Administration

Contact persons tab:
Screenshot 2024-09-09 at 13-21-57 Editing indicator - Climate - Administration

Reporting tab:
Screenshot 2024-09-09 at 13-22-05 Editing indicator - Climate - Administration

Relationships tab:
Screenshot 2024-09-09 at 13-22-17 Editing indicator - Climate - Administration

indicators/models.py
```diff
@@ -440,6 +440,23 @@ class Indicator(ClusterableModel, index.Indexed, ModificationTracking, PlanDefau
         features=['link'],
     )
 
+    show_trendline = models.BooleanField(
+        default=True, verbose_name=_('show trendline'),
+    )
+
+    desired_trend = models.CharField(
+        blank=True, null=True, verbose_name=_('desired trend'), max_length=20, default=None,
```
bbliem: Is there a reason for allowing NULL here? In general, it's preferable to use empty strings in char fields, as mentioned in the [Django documentation](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null), unless you have a good reason for not doing so.

indicators/models.py
```diff
@@ -440,6 +440,23 @@ class Indicator(ClusterableModel, index.Indexed, ModificationTracking, PlanDefau
...
+        choices=(
+            ('increase', _('increase')),
+            ('decrease', _('decrease')),
+            (None, _('attempt to detect automatically')),
+        ),
+        help_text=_(
+            "Is improvement indicated by an increasing or decreasing trend in the numerical values of this "
+            "indicator's goals?",
```
bbliem: The wording could be improved. Perhaps using the active voice makes it clearer.

indicators/tests/factories.py
```diff
@@ -60,6 +60,8 @@ class Meta:
     description = "Indicator description"
     min_value = 0.0
     max_value = 100.0
+    show_trendline = False
+    desired_trend = "decrease"
```
bbliem: Minor detail: Personally I prefer single quotes for text that is not shown to the user but just used internally, like this one. Other people might have different preferences / code styles, but I've tried to keep it like this at least in code that I touched.

indicators/wagtail_admin.py
```diff
@@ -382,123 +383,186 @@ class IndicatorAdmin(AplansModelAdmin):
...
-        request = ctx_request.get()
-        instance = cast(Indicator, ctx_instance.get())  # FIXME: Fails when creating a new indicator
-        basic_panels = list(self.basic_panels)
-        advanced_panels = list(self.advanced_panels)
+    def _get_basic_information_tab(self, instance, request) -> ObjectList:
+        """Get basic information tab for edit view."""
+        panels: list[Panel] = []
         plan = request.user.get_active_admin_plan()
+        is_general_admin = request.user.is_general_admin_for_plan(plan)
+        is_linked_to_common_indicator = instance and instance.common
```
bbliem: The name of this variable is misleading because `instance and instance.common` is not a boolean value. (It is `None` or a `CommonIndicator` because of the behavior of `and`.) It may be better to change the value to `bool(instance and instance.common)`. Or keep it as it is but call the variable `linked_common_indicator`.

indicators/wagtail_admin.py
```diff
@@ -382,123 +383,186 @@ class IndicatorAdmin(AplansModelAdmin):
...
-        advanced_panels.insert(
-            2, FieldPanel('common', widget=autocomplete.ModelSelect2(url='common-indicator-autocomplete')),
+        # Further information
+        panels.append(
+            MultiFieldPanel(
+                children=[
+                    CustomizableBuiltInFieldPanel('description'),
+                    CustomizableBuiltInFieldPanel('reference'),
+                ],
+                heading=_('Further information'),
```
bbliem: This one, for example, I'd write with double quotes personally. (See comment above.) As I said, it's a matter of taste.

indicators/wagtail_admin.py
bbliem: Yeah sounds good.

bbliem (approved):
Nice!

indicators/models.py
MWargelin: No reason, just a brainfart. Good catch, thanks!

indicators/models.py
MWargelin: Yeah, this was really difficult to phrase. I changed it to "Which trend in the numerical values of this indicator's goals indicates improvement: when the values are increasing or decreasing?", what do you think of that?

indicators/tests/factories.py
MWargelin: Ah, I've interpreted the [PEP8 recommendation](https://peps.python.org/pep-0008/#string-quotes) as "pick either single quotes or double quotes and stick with it", but I guess the rule to follow can be more complicated than that. Is everyone on the same page with this? I feel like unless this is a collective agreement we just end up with things being messy and single quotes not meaning anything because not everyone follows the rule of using those for internal strings.

In general I like double quotes because with English language you more often end up having to use single quotes than double quotes in the string value itself, so using double quotes lets you more often not having to use any escape characters.

Anyway, changed this now to single quotes.

indicators/wagtail_admin.py
MWargelin: Whoops, I've always thought `and` always converts the value to boolean automatically, but you're right. Good to know, thanks! I changed this to `bool(instance and instance.common)`

indicators/wagtail_admin.py
MWargelin: Changed this to double quotes, more thoughts in the other comment

PR #253, opened 2024-09-05
Fix/indicator value race condition

tituomin:


Bhavatu (approved):
Should do the trick.

tituomin:
Merged manually to main.

PR #263, opened 2024-09-18
Fix the handling of rejections in workflows

tituomin:
In Wagtail, the property current_workflow_task returns None if the workflow is not in progress. We actually need to differentiate between the states where a specific workflow task has been rejected, and the model instance has been returned back to the previous stage, and handle that case explicitly.


Bhavatu (approved):
Looks logical, didn't test though but i trust you did. Notice the ruffs

bbliem (commented):
The docstring doesn't help me understand what this method is supposed to do. But I trust you tested it successfully.

PR #268, opened 2024-09-27
Fix/budget migrations

tituomin:
The way I had implemented the default= value for Dataset.schema was not a good idea; it broke the migrations which are run with an old version of the DatasetSchema db schema.


tituomin:
@Bhavatu Symptom: when creating a new database, the migrations were run from 0000... until 0002 which broke because the db  schema of the current DatasetSchema was incompatible with the db schema after migration 0001

Bhavatu (approved):
LGTM

PR #276, opened 2024-10-09
Indicator schema: Related plan indicators

Bhavatu:
Resolve related plan indicators:
Asana: https://app.asana.com/0/1206017643443542/1208507870598008


indicators/schema.py
```diff
@@ -377,6 +399,35 @@ def resolve_indicator(self, info, restrict_to_publicly_visible: bool, **kwargs):
...
+        # FIXME: This is sucky, maybe convert Category to a proper tree model?
+        f = (
+            Q(id=category) |
+            Q(parent=category) |
+            Q(parent__parent=category) |
+            Q(parent__parent__parent=category) |
+            Q(parent__parent__parent__parent=category)
+        )
+        descendant_cats = Category.objects.filter(f)
+        qs = qs.filter(categories__in=descendant_cats).distinct()
```
tituomin: I am left wondering how this category filtering would work, since categories are plan-specific and here we are querying indicators across different plans. Am I missing something?

Bhavatu: You're right that in the end the filtered indicators would all be part of the same category and plan, if you supply the resolver with a category to filter by. Or was that what you meant?

tituomin (approved):

PR #277, opened 2024-10-10
Fix the most pressing performance degradation in the Admin people index view

tituomin:
Locally, before this fix:
⛁ 655 SQL queries took 10018 ms /admin/people/person/

Locally, after the fix: (cold cache)
⛁ 184 SQL queries took 360 ms /admin/people/person/

Locally, after the fix: (warm cache)
⛁ 160 SQL queries took 247 ms /admin/people/person/

Explanation: the biggest performance hit comes from calling `user.can_access_admin()` in a loop. While it would be nice to use that function because it actually calls the actual implementation that is used when determining user access, it is very difficult to cache this result the way it's currently implemented. I have decided to add another implementation tuned for mass querying and caching, while writing some tests which hopefully give errors when these two implementations get out-of-sync. It's not ideal but the new implementation is so simple it's hopefully ok.

---
  - https://app.asana.com/0/0/1208110855280844

### DIFF
--- a/actions/perms.py
+++ b/actions/perms.py
@@ -1,11 +1,14 @@
+from __future__ import annotations
+
 from functools import lru_cache
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
 from django.contrib.contenttypes.models import ContentType
 from django.db.models.query import QuerySet
-from treelib import Tree
 from wagtail.models import PAGE_PERMISSION_TYPES, GroupPagePermission
+
+from treelib import Tree
 
 from content.models import SiteGeneralContent
 from indicators.models import (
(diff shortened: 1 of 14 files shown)


actions/perms.py
```diff
@@ -362,3 +365,99 @@ def has_delete_permission(self, request, obj=None):
...
+) -> set[int]:
+    """
+    Return a set of Person pks specifying all the persons who do not have any edit rights in the system.
+
+    Receive a snapshot with sets of pks for all model instances in the db which are needed for determining
+    if a person has some edit rights anywhere in the system to determine this.
+
+    """
+    persons_with_permissions_pks = (
+        superusers |
```
Bhavatu, line 429: Is the docstring correct? "specifying all the persons who do **not** have any edit rights in the system"

Bhavatu (approved):
Looks good, didn't notice anything obviously wrong while reading through. Good amount of tests :+1:

actions/perms.py
tituomin: Good catch. It used to be the other way around.

PR #278, opened 2024-10-10
Fix Excel export when actions have no implementation phase

bbliem:
See [discussion in Slack](https://kausaltech.slack.com/archives/C06T7CU6XC0/p1728568818051319).

### DIFF
--- a/reports/spreadsheets/excel_report.py
+++ b/reports/spreadsheets/excel_report.py
@@ -425,11 +425,10 @@ class ExcelReport:
             raise ValueError('Only one or two dimensional pivot tables supported')
         action_df = action_df.fill_null('[' + _('Unknown') + ']')
         if len(labels) == 1:
-            return action_df\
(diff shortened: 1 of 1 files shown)


reports/spreadsheets/excel_report.py
```diff
@@ -425,11 +425,10 @@ def _get_aggregates(self, labels: tuple[str], action_df: polars.DataFrame):
             raise ValueError('Only one or two dimensional pivot tables supported')
         action_df = action_df.fill_null('[' + _('Unknown') + ']')
         if len(labels) == 1:
-            return action_df\
-                .groupby(labels)\
-                .count()\
-                .sort(reversed(labels), descending=False)\
```
bbliem, line 431: What's the point of sorting (and also reversing `labels`) when we know that `len(labels) == 1`?

tituomin (approved):

PR #280, opened 2024-10-11
Make reports more robust with missing values and columns

tituomin:
When the reporting code triest to generate pivot tables from the existing data, it should skip trying if either the column does not exist at all in the data, or it does exist but all the values are missing.

### DIFF
--- a/reports/spreadsheets/excel_report.py
+++ b/reports/spreadsheets/excel_report.py
@@ -454,9 +454,6 @@ class ExcelReport:
                 'type': 'column',
             },
         ]
-        # Work around problems if none of the actions have an implementation phase and that column would have type null
-        action_df = action_df.with_columns(polars.col(_('Implementation phase')).cast(polars.String))
-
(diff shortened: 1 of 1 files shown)


reports/spreadsheets/excel_report.py
```diff
@@ -471,6 +468,18 @@ def post_process(self, action_df: polars.DataFrame):
...
+                # Do not generate the pivot sheet if there are no values for
+                # some of the required fields
+                if field_label not in action_df:
+                    skip = True
+                    break
+                if action_df.get_column(field_label).dtype.is_(polars.datatypes.Null):
+                    skip = True
+                    break
+            if skip:
+                continue
```
bbliem: Nice. I think it could be even nicer with something like:
```python
def skip(field_label):
    return field_label not in action_df or action_df.get_column(field_label).dtype.is_(polars.datatypes.Null)

if any(skip(field_label) for field_label in grouping):
    continue
```

bbliem (approved):

PR #281, opened 2024-10-18
Contactform comment and email field settings

Bhavatu:
Comment (Feedback textfield) and Email now have settings "required", "optional" and "excluded" for the UI.

Comment field is now nullable.

Related UI PR: https://github.com/kausaltech/kausal-watch-ui/pull/420


actions/blocks/action_content.py
```diff
@@ -282,10 +282,29 @@ class Meta:
...
 
 
+
+
 class BaseContactFormBlock(blocks.StructBlock):
+    FEEDBACK_FIELD_CHOICES = [
+    ('required', _('Required')),
+    ('optional', _('Optional')),
+    ('excluded', _('Excluded')),
+]
```
bbliem: The indenting is messed up here.

actions/blocks/action_content.py
```diff
@@ -282,10 +282,29 @@ class Meta:
     ]
 
 
+
+
```
bbliem: Two empty lines is enough.

actions/blocks/action_content.py
```diff
@@ -282,10 +282,29 @@ class Meta:
...
+    ('required', _('Required')),
+    ('optional', _('Optional')),
+    ('excluded', _('Excluded')),
+]
     heading = blocks.CharBlock(required=False, default="", label=_('Heading'))
     description = blocks.CharBlock(required=False, default="", label=_('Description'))
-    email_required = blocks.BooleanBlock(required=False, default=True, label=_('Email required') )
+    feedback_setting = blocks.ChoiceBlock(
+        choices=FEEDBACK_FIELD_CHOICES,
+        default='required',
```
bbliem: Specifying the default here as a string ("required") could be trouble if for some reason the strings representing the different choices in `FEEDBACK_FIELD_CHOICES` change because we might forget updating the default here. I haven't checked if it works for blocks, but for model fields one way to avoid this problem is the following (copied from `aplans/utils.py`):
```python
class RestrictedVisibilityModel(models.Model):
    class VisibilityState(models.TextChoices):
        INTERNAL = 'internal', _('Internal')
        PUBLIC = 'public', _('Public')

    visibility = models.CharField(
        blank=False, null=False,
        default=VisibilityState.PUBLIC,
        choices=VisibilityState.choices,
        max_length=20,
        verbose_name=_('visibility'),
    )
```
Can something like this be used here as well?

actions/blocks/action_content.py
bbliem: Also, I'm a bit concerned that mixing visibility and "optionality" into one thing is not very clean. Couldn't we just have one Boolean (i.e., a checkbox) for "required", and one for "visible" (or "hidden" or "excluded" or whatever you want to call it)?

The fact that it's not clean to mix these two is reflected in the fact that the current draft calls these fields "feedback setting" and "email setting", which is very vague. You can't understand what "feedback setting" or "email setting" is supposed to mean without looking at the definition of the fields. I believe it ideally should be clear from every field name alone what the field means.

actions/blocks/action_content.py
```diff
@@ -282,10 +282,29 @@ class Meta:
...
+        label=_('Feedback field setting'),
+        help_text=_('Configure the feedback field visibility and requirement.'),
```
bbliem: As mentioned above, I'm not sure if mixing these two is a good idea. Also, I believe the word "requirement" is a bit confusing here and this should perhaps rephrased, but I don't have a specific alternative ready at the moment.

actions/blocks/action_content.py
```diff
@@ -282,10 +282,29 @@ class Meta:
...
+    )
+
+    email_setting = blocks.ChoiceBlock(
+        choices=FEEDBACK_FIELD_CHOICES,
```
bbliem: The thing about mixing things also applies here. If we leave it like this, however, I think it's not a good idea to use `FEEDBACK_FIELD_CHOICES` for something that has nothing to do with the feedback field. Either rename `FEEDBACK_FIELD_CHOICES` to something generic or copy it and call it `EMAIL_FIELD_CHOICES` or so.

feedback/models.py
```diff
@@ -35,7 +35,7 @@ class FeedbackType(models.TextChoices):
 
     name = models.CharField(max_length=100, null=True, blank=True, verbose_name=_("name"))
     email = models.EmailField(null=True, blank=True, verbose_name=_("email address"))
-    comment = models.TextField(verbose_name=_("comment"))
+    comment = models.TextField(verbose_name=_("comment"), blank=True, null=True)
```
bbliem: Is there a reason for `null=True`? In general, setting both to true at the same time is discouraged. I quote the Django documentation:

> Avoid using [null](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null) on string-based fields such as [CharField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField) and [TextField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.TextField). If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a [CharField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField) has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

So unless there are issues with unique constraints, it would be better to omit `null=True`.

bbliem (dismissed):

actions/blocks/action_content.py
Bhavatu: That was my first thought how to do it, but then i pondered about the practicality about two booleans. What if user sets both to true, then the visibility overrides the required boolean. But still seems clunky to have two booleans to me, when hiding the field effectively makes it not required anyways.

feedback/wagtail_admin.py
```diff
@@ -58,6 +58,12 @@ class UserFeedbackInspectView(InspectView):
     # (and the whole class) can be deleted
     any_permission_required = ["add", "change", "delete", "view"]
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        for field in context['fields']:
+            if field['value'] == '' or field['value'] is None:
+                field['value'] = 'None'
```
tituomin, line 65: So this is a user-visible value in place of an empty value. Maybe at some point we would want to use something like `_('empty')` or `_('not provided')` here? But not urgent right now.

tituomin (approved):

PR #287, opened 2024-11-01
[WIP] Plan visibility

Bhavatu:
Visibility setting for plans. Depending on `published_at`, resolvers return content related to plan only if logged in to public UI.

---

**This is a draft, in the process of being finished.**


actions/schema.py
```diff
@@ -1633,6 +1648,8 @@ def resolve_action(
             return action
 
         user = info.context.user
+        if not plan_obj.is_visible_for_user(user):
```
tituomin, line 1795: I think it might be cleaner and less error prone if we included in `Action.is_visible_for_user` a check for `action.plan.is_visible_for_user`. Because I'm pretty sure we want the plan.visible... == False to always override action.visible... == True

tituomin: But to expand on that idea, I do wonder if we might want to add to all models with potentially a restricted visibility based on their plan an implementation of `visible_for_user` which would just delegate to `plan.visible_for_user`

tituomin (commented):
Looks clean, but see my comments about maybe moving code to the models themselves.
I also have am not quite certain that all possible ways to get to the plan from GQL are covered. I'll do some more checking.

actions/schema.py
tituomin: I don't know, but if we had a uniform API for handling visibility (regardless of whether all the models have a visibility field themselves), maybe we could more easily verify we are using it consistently?

tituomin (dismissed):
Add a feature flag.
Also, I only just now noticed that actually the viewing permissions should be based on (plan, user) pair, meaning that users in plan X should not see restricted plan Y.

```   def construct_perm_q(self, user: User, action: ObjectSpecificAction) -> Q | None:
        """
        Construct permission query for authenticated users.
        Allow viewing all plans but require specific permissions for other actions.
        """
        if action == 'view':
            return Q()  # Can view all plans
        # Add other permission checks for change/delete if needed
        return None
```

MWargelin:
Looks like this is in danger of getting rotten and forgotten, some confilicts are starting to appear. Is this still in radar @Bhavatu @tituomin ?

tituomin:
> Looks like this is in danger of getting rotten and forgotten, some confilicts are starting to appear. Is this still in radar @Bhavatu @tituomin ?

Thanks for the reminder! Yes, I have actually rebased this recently and am working on it. However, something more urgent came up. I'm thinking I might still want to push the rebased version. Basically I have taken over the task.

MWargelin:
> > Looks like this is in danger of getting rotten and forgotten, some confilicts are starting to appear. Is this still in radar @Bhavatu @tituomin ?
> 
> Thanks for the reminder! Yes, I have actually rebased this recently and am working on it. However, something more urgent came up. I'm thinking I might still want to push the rebased version. Basically I have taken over the task.

Cool, I got worried noticing this is couple of months old PR already. But great that it is in control :+1:

tituomin:
This branch has been rebased and reviewed and merged into master. See commits 05d8eadb to 45153da2

PR #289, opened 2024-11-04
fix common indicator creation

Bhavatu:
https://sentry.kausal.tech/organizations/kausal/issues/2816/?project=3


bbliem (commented):
If there is no `pk`, then the object is being created but `save()` has not been called yet. If it works now and if you think that it's fine to do those checks only for instances that are already in the database (as opposed to instances that are about to be created), then it's probably fine.

PR #290, opened 2024-11-05
Show active plan name on sidebar

MWargelin:
Shows the active plan name on the sidebar.

This is preparation for the feature to copy a plan, to make sure users keep track of which plan they are modifying (so that they don't accidentally edit the copy when they mean to edit the original).

This might not be the prettiest solution or the one that makes it most clear to the users which plan is being worked on, so this might need some tuning. But at least the mechanism and the hooks to style it with CSS are discovered and in place, so future development should be easy.

### DIFF
--- a/admin_site/static/css/admin-styles.css
+++ b/admin_site/static/css/admin-styles.css
@@ -43,6 +43,10 @@
   padding: 10px 0;
 }
 
+.sidebar--slim .active-plan {
+  display: none;
+}
+
 .login .login-logo {
   width: 3.5rem;
 }
--- a/admin_site/templates/wagtailadmin/base.html
+++ b/admin_site/templates/wagtailadmin/base.html
@@ -15,10 +15,10 @@
 {% block branding_logo %}
     {% if active_client and active_client.logo %}
         {% image active_client.logo width-80 %}
-        <!-- <h5>{{ active_plan.name }}</h5> -->
     {% else %}
         <img src="{% static 'images/kausal-avatar-dark-green.svg' %}" alt="Kausal Logo" width="80" />
     {% endif %}
+    <div class="active-plan">{{ active_plan.name }}</div>
 {% endblock %}
 
 {% block extra_css %}


tituomin:
I will test this locally a bit.

MWargelin:
> I will test this locally a bit.

@tituomin By the way I had a hard time making Firefox to use the updated CSS style sheet instead of one it had cached and was wondering for ages why my modifications do nothing. But finally I found out for Firefox ctrl+F5 forces reloading the style sheet, so had to do that to see the plan name disappear when collapsing the sidebar.

tituomin:
Could you rebase to main?

tituomin:
There used to be an automatic checksum appended to the static files, wonder why not anymore. That should help with the updates. Or was it only in production, I could check

tituomin (approved):
This is good for now! I do wonder if in some cases this info might be important enough to find a way to show even when menu is narrowed. Out of scope for this one though.

tituomin:
@MWargelin No it was just for production, the cache-busting thing. Well at least it should then immediately refresh for customers https://github.com/kausaltech/kausal-watch-private/commit/b44875041347235455637b1cdc0226f89232c51d

MWargelin:
replies @tituomin 

> @MWargelin No it was just for production, the cache-busting thing. Well at least it should then immediately refresh for customers [b448750](https://github.com/kausaltech/kausal-watch-private/commit/b44875041347235455637b1cdc0226f89232c51d)

Alright, good to know!

> This is good for now! I do wonder if in some cases this info might be important enough to find a way to show even when menu is narrowed. Out of scope for this one though.

Yeah, I was thinking that, but also I couldn't think of a way that that much text would look good on a sidebar that narrow. The overflowing buggy looking text (which happens if the CSS rule to hide the plan name from collapsed sidebar is left out completely) actually works quite well in the sense that the plan name is always visible, it just looks so much like a bug that leaving that in would make the admin UI look pretty cheap :sweat_smile: 

> Could you rebase to main?

Used the rebase and merge option when merged the PR :+1:

PR #292, opened 2024-11-07
Fix read only fields bugs

tituomin:
For the recent read_only field support for attributes there was a critical bug which caused a crash for attribute types without help text.

This PR includes other fixes as well.

### DIFF
--- a/actions/attributes.py
+++ b/actions/attributes.py
@@ -59,7 +59,10 @@ class AttributeFieldPanel[M: models.ModelWithAttributes](FieldPanel[M]):
             ):
                 rev = self.instance.get_latest_revision()
                 draft_attributes = DraftAttributes.from_revision_content(rev.content.get('attributes'))
-                attribute_value = draft_attributes.get_value_for_attribute_type(self.panel.attribute_type)
+                try:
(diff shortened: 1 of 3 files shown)


actions/attributes.py
```diff
@@ -597,14 +603,26 @@ def get_form_fields(
             choice_options, initial=initial_choice, required=False, help_text=self.instance.help_text_i18n,
         )
         is_public = self.instance.instances_visible_for == self.instance.VisibleFor.PUBLIC
+        if not editable:
+            return [FormField(
```
Bhavatu: Is it ok to return this already here? We don't need the fields for all the languages? (https://github.com/kausaltech/kausal-watch-private/pull/292/files?diff=split&w=1#diff-8c8cc74d10aebd2123804c9e21214402fc4af802859c9d7ba02a390129630272R629)

tituomin: Good point, I will see how this behaves in a multi-language plan.

tituomin: Yes indeed I had forgot to implement multilingual support at all, so thanks, well spotted! Fixed in 22dbe387d2fea2f6ef61ac43e21715a78446cac9

PR #293, opened 2024-11-14
Feature/copy plan

bbliem:
Provides the management command `copy_plan`.

Right now the tests are failing. This is because I needed to add some through models; for example, for the `schedule` relationship of `Action`. The `get_relation_kwargs()` function of DRF then pops the `'queryset'` key from the kwargs, which causes an error in our code. I'll investigate that, but I suppose the copying functionality can be tested and reviewed already.


copying/main.py
```diff
@@ -0,0 +1,496 @@
...
+    'general_content': {},
+    'impact_groups': {},
+    'indicator_levels': {},  # We don't copy the indicators themselves
+    'notification_settings': {},
+    'plan_common_category_types_through': {},
+    'plan_related_organizations_through': {},
+    'public_site_viewers': {},
+    'report_types': {
+        'reports': {
+            'action_snapshots': {},  # FIXME: copy `action_version` otherwise wrong action is linked to snapshot
```
tituomin: Is this FIXME still unresolved?

copying/management/commands/copy_plan.py
```diff
@@ -0,0 +1,67 @@
...
+            help="Choose a hostname for the new plan's site",
+        )
+        parser.add_argument(
+            '--dest-identifier',
+            help="Use the given value as the copy's identifier (default: Append '-copy')",
+        )
+        parser.add_argument(
+            '--dest-name',
+            help="Use the given value as the copy's name (default: Append a string containing the current date)",
+        )
```
tituomin, line 26: Should we add an option for setting the version_name of the new Plan?
https://github.com/kausaltech/kausal-watch-private/blob/main/actions/models/plan.py#L207-L210
It's a bit more structured than name suffix below and is already used in related features

copying/management/commands/copy_plan.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'identifier',
+            help="Copy the plan with the given identifier",
+        )
+        parser.add_argument(
+            'hostname',
+            help="Choose a hostname for the new plan's site",
+        )
```
tituomin: Not sure right away how the hostnae here works, but should we add an option for setting the basepath of a plan domain for the copy?

tituomin (approved):
I like how cleanly this is split into logical parts that do as little as possible, very easy to undestand. Nice! Haven't yet tested running it.
My requests are more like additional requirements, and can also be implemented separately and left out of this PR.

copying/management/commands/copy_plan.py
bbliem: Good idea. Are you suggesting we get rid of the name suffix by default? So the name should be the same as in the original by default?

copying/main.py
bbliem: Yes, but it might not be a bad thing. :thinking:

copying/management/commands/copy_plan.py
bbliem: This one only refers to the `Site` model. But you're right that we should do something about the `PlanDomain` instances. At the moment, we just copy them without modifications, but this of course is not good.

By the way, this made me curious why we don't run into integrity violations because the combination `(hostname, base_path)`  should be unique. The problem is that `base_path` may be NULL, and NULL is always considered different from NULL. (Sometimes this is what we want, when we want uniqueness only if a nonempty value is set.) Indeed, for some reason, we set `base_path` to NULL instead of the empty string.

When trying to copy a plan that actually has a domain with a nonempty base path, we indeed get a uniqueness violation.

I believe the best course of action when copying a plan is to not copy any domains and have an admin create whatever domains they want -- if any -- themselves.

Besides, I wonder why we allow NULL for `base_path` anyway because in this case we want to enforce uniqueness also when there is no base path. So perhaps we should make this field non-nullable and use the empty string, as is the preferred way with Django.

Bhavatu:
I am reading this through now, note that ruff is barking and CI tests are failing.

copying/views.py
```diff
@@ -0,0 +1,145 @@
...
+                'label': self.get_page_title(),
+                'sublabel': self.get_page_subtitle(),
+            }
+        )
+        return items
+
+    def form_valid(self, form) -> HttpResponse:
+        logger.info(f"Queueing task for copying plan {self.plan_id}")
+        assert isinstance(copy_plan, DjangoTask)
+        copy_plan.delay_on_commit(
```
Bhavatu, line 127: I didn't notice anything that hinders an impatient user from spamming the copy task, e.g. a redis key that would be set when starting and then unset when task is ready.

Bhavatu (approved):
Preliminary approval, didnt notice anything obviously wrong. just see my minor comment and fix the failing tests and ruff.

Great work. I appreciate the clean code and documentation, made it much easier to understand.

PR #295, opened 2024-11-18
Add possibility to export feedback as spreadsheet

MWargelin:
Add possibility to export user feedback as CSV or XLSX files (to be further processed in e.g. Excel).

The export is found behind the three dots in the feedback listing view:
image

### DIFF
--- a/feedback/wagtail_admin.py
+++ b/feedback/wagtail_admin.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+from collections import OrderedDict
 from typing import TYPE_CHECKING
 
 from django.contrib.admin.utils import quote
 from django.urls import path, reverse
 from django.utils.translation import gettext_lazy as _
 from wagtail.admin.ui.tables import BooleanColumn
+from wagtail.coreutils import multigetattr
 from wagtail.permission_policies.base import ModelPermissionPolicy
 from wagtail.snippets import widgets as wagtailsnippets_widgets
 from wagtail.snippets.models import register_snippet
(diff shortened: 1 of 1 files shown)


feedback/wagtail_admin.py
```diff
@@ -66,12 +68,84 @@ def get_context_data(self, **kwargs):
...
+        """Get a list of all user-defined additional fields of the feedback form present in the queryset."""
+        if self.additional_fields_cache is not None:
+            return self.additional_fields_cache
+
+        additional_fields = []
+        for feedback in self.get_queryset():
+            if feedback.additional_fields is not None:
+                additional_fields += feedback.additional_fields.keys()
+
+        duplicates_removed = list(dict.fromkeys(additional_fields))
```
bbliem, line 121: I suppose you're using `list(dict.fromkeys(...))` because you want to eliminate duplicates without changing the order. If order weren't an issue, I would have said that it's a terribly clumsy way of eliminating duplicates. But with order being significant, it seems that this is the most straightforward way of doing this since Python dicts started being ordered.

"Python encourages you to write code that's easy to read", they said. What a piece of bullshit!

feedback/wagtail_admin.py
```diff
@@ -66,12 +68,84 @@ def get_context_data(self, **kwargs):
...
+        for field in self.list_export:
+            try:
+                value = multigetattr(item, field)
+            except AttributeError:
+                # If the field is not an attribute of the feedback model, check if it's one of all additional fields.
+                if field not in self._get_additional_fields():
+                    raise
+
+                # The field might still not exist in this particular item's additional_fields. Use N/A as a fallback.
+                value = 'N/A'
```
bbliem: Maybe 'N/A' should be translatable. But actually why not use an empty string or so?

MWargelin: Empty string and 'N/A' actually have different semantics now in the exported spreadsheet. I tried to differentiate the situations where

1. the user never had a chance to answer to the question because the field wasn't in the feedback form when they answered ('N/A')
2. The user chose not to write an answer to the field that was there (empty string)

feedback/wagtail_admin.py
MWargelin: Yeah I didn't like this either but Python doesn't seem to have an ordered set and I didn't find a better way to make that kind of behavior happen. I think the ordering matters because the feedback form questions might refer to each other and be more sensible in certain order, I found some examples such as

- Do you oppose this action?
- If you oppose, why?

(Python had a weak moment here but I think mostly it is really good in enabling writing easy to read code :stuck_out_tongue: )

feedback/wagtail_admin.py
MWargelin: I'm also thinking maybe it shouldn't be translated? `N/A` is pretty standard way to mark missing information when doing data wrangling, maybe it should be kept that way in every language. Or what do you think?

bbliem: It may be known to data wranglers anywhere in the world, but maybe not to the general population or public administration employees. As you know, "N/A" stands for "not available" or similar expressions, but they are all English. Maybe somewhere people have an aversion to the English language and thus not use "N/A"? Or maybe they have a local expression that is commonly used for such cases? Who knows how it is somewhere in, say, southeast Asia, African languages, or... French!?

MWargelin: Alright, changed it now to translatable string :+1:

PR #296, opened 2024-11-18
Add registry for configuring Wagtail blocks based on Action fields

tituomin:
We have several different contexts where Wagtail blocks representing Action fields are used. (Action details page, action dashboard list, reports; also filter blocks which are not part of this PR)

Previously the different blocks used in the different contexts were added in an ad hoc way. Some block classes were generated in a disorganized way while others were manually specified even though part of them might easily have been generated as well.

Here we implement a new registry for model fields. It can be used to explicitly configure the field-derived blocks for a model, now only for Actions. We also introduce some checks, for example we check that if a new public field is added to an Action, there is also an explicit block configuration for it.

One goal for this refactoring is to make sure that for all Action dashboard list column blocks a corresponding Action report block exists. (This is verified in the tests but currently only a warning is emitted.) This is because we will no longer use a UI implementation for the UI dashboard excel export, and will reuse the reports app excel export code. Thus, all dashboard columns need to be exportable.

In the future, we should reduce the amount of different block types further -- for example the Dashboard column blocks are very similar to the other block types, only with a different naming convention for the fields.


tituomin:
The mypy errors were there already in the old files and appear now just because I moved code around. I will probably not be fixing those now.

actions/blocks/action_content_blocks.py
```diff
@@ -0,0 +1,250 @@
...
+    graphql_interfaces = (FieldBlockMetaInterface, )
+
+    class Meta:
+        label = _("Indicator Causal Chain")
+
+
+class BaseDatasetsBlock(blocks.StructBlock):
+    heading = blocks.CharBlock(
+        required=False,
+        help_text=_("What heading should be used in the public UI for the Dataset?"),
```
bbliem: Dataset should not be upper case.

actions/tests/test_action_field_blocks.py
```diff
@@ -0,0 +1,75 @@
...
+        msg = (f'No blocks configured for public field "{field.name}" of Action. '
+               'Please configure or disable blocks for that field in actions.action_fields.py')
+        warnings.warn(msg, stacklevel=1)  # TODO: make this a test failure
+
+    default_field_type = get_field_type(field)
+    if field_props.is_disabled():
+        return True
+    assert \
+        field_props.field_type in ('custom', default_field_type), \
+        f'Wrong type for field {field.name}'
```
bbliem: Probably it's a matter of taste, but I"m not a fan of escaped newlines in Python and would rather wrap the expression in parentheses.

actions/tests/test_action_field_blocks.py
```diff
@@ -0,0 +1,75 @@
...
+from actions.models import Action
+
+
+def get_field_type(field):
+    if isinstance(field, ForeignKey):
+        return 'single'
+    if isinstance(field, ManyToOneRel):
+        return 'many'
+    if isinstance(field, ManyToManyField):
+        return 'many'
```
bbliem: Not entirely sure, but could we check for the attributes `one_to_one`, `one_to_many`, `many_to_many` or `many_to_one` in `field`. Django does this internally in many cases. Perhaps this is more robust than type checking.

I'm also a bit unsure about whether setting this to "many" for `ManyToOneRel` because of this comment in the Django code:

>    Note: Because we somewhat abuse the Rel objects by using them as reverse
>    fields we get the funny situation where
>    ``ManyToOneRel.many_to_one == False`` and
>    ``ManyToOneRel.one_to_many == True``. This is unfortunate but the actual
>    ManyToOneRel class is a private API and there is work underway to turn
>    reverse relations into actual fields.

actions/tests/test_action_field_blocks.py
```diff
@@ -0,0 +1,75 @@
...
+
+    """
+
+    public_fields = Action.public_fields
+    # for field in Action._meta.get_fields():
+    if field.name not in public_fields:
+        # Blocks for non-public fields are not allowed
+        with pytest.raises(KeyError):
+            action_registry[field.name]
+        return True
```
bbliem: Does returning a Boolean do anything in tests?

aplans/tests/test_field_registry.py
```diff
@@ -0,0 +1,77 @@
...
+from reports import report_formatters
+
+pytestmark = pytest.mark.django_db
+
+
+@pytest.fixture()
+def field_name():
+    seq = 1
+    def _field_name():
+        return f'field-{seq!s}'
```
bbliem: Why the `!s` suffix? It should be equivalent to use `{seq}` since string interpolation in f-strings converts stuff inside `{...}` to string anyway.

bbliem (approved):

actions/tests/test_action_field_blocks.py
tituomin: Yes, it was because assert is a statement with special syntax, and I couldn't just break it like
```
assert (
    condition_expression,
    message
)
```

However, I do now realize I can put the parentheses only around the message part, maybe that's what you meant? I'll do it that way.

actions/tests/test_action_field_blocks.py
tituomin: Great recommendation, I tested it and it seems to work really nicely.

actions/tests/test_action_field_blocks.py
tituomin: No, fixing.

aplans/tests/test_field_registry.py
tituomin: I added that because my old version of ruff was somehow demanding that. I can't reproduce it anymore, so will remove.

PR #297, opened 2024-11-21
Fix/clean types

tituomin:


users/models.py
```diff
@@ -57,6 +57,7 @@ class User(AbstractUser):
 
     auth_token: Token
     person: Person
+    wagtail_userprofile: UserProfile
```
juyrjola: This one should probably be:
```
wagtail_userprofile: RevOne[User, UserProfile]
```

juyrjola (commented):
Beside that one change, LGTM!

PR #298, opened 2024-11-26
Add active plan status to side bar

MWargelin:
Adds status information about the active plan to the status bar.

The design created for this feature can be found here: https://app.asana.com/0/1206017643443542/1208607227000757/f

### DIFF
--- a/admin_site/templates/wagtailadmin/base.html
+++ b/admin_site/templates/wagtailadmin/base.html
@@ -18,7 +18,30 @@
     {% else %}
         <img src="{% static 'images/kausal-avatar-dark-green.svg' %}" alt="Kausal Logo" width="80" />
     {% endif %}
-    <div class="active-plan">{{ active_plan.name }}</div>
+
+    <div class="active-plan-info">
+        <span class="active-plan-name">{{ active_plan.name_i18n }}</span>
+        <div class="active-plan-status">
+            <!-- TODO: Enable this once locking plans is implemented-->
+            <!-- {% if active_plan.is_locked %}
+                <span class="active-plan-locked" aria-label="{% trans 'locked plan' %}">{% icon name="lock" %}</span>
+            {% endif %} -->
+
+            {% if active_plan.version_name %}
+                <span class="active-plan-version-name">{{ active_plan.version_name }}</span>
+            {% endif %}
+
+            <div>
+                <span class="active-plan-live {% if active_plan.is_live %}public{% else %}internal{% endif %}">
+                    {% if active_plan.is_live %}
+                        {% trans "Public" %}
+                    {% else %}
+                        {% trans "Internal" %}
+                    {% endif %}
+                </span>
+            </div>
+        </div>
+    </div>
 {% endblock %}
 
 {% block extra_css %}
--- a/actions/models/plan.py
+++ b/actions/models/plan.py
@@ -649,7 +649,8 @@ class Plan(ClusterableModel, ModelWithPrimaryLanguage):
         return primary_root_page
 
     def is_live(self):
-        return self.published_at is not None and self.archived_at is None
+        now = self.now_in_local_timezone()
+        return self.published_at is not None and self.published_at <= now and self.archived_at is None
 
     def get_optional_locale_prefix(self, locale: str):
         if locale.lower() == self.primary_language.lower():
--- a/admin_site/static/css/admin-styles.css
+++ b/admin_site/static/css/admin-styles.css
@@ -43,7 +43,55 @@
   padding: 10px 0;
 }
 
-.sidebar--slim .active-plan {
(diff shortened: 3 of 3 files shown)


admin_site/templates/wagtailadmin/base.html
```diff
@@ -18,7 +18,23 @@
...
         <img src="{% static 'images/kausal-avatar-dark-green.svg' %}" alt="Kausal Logo" width="80" />
     {% endif %}
-    <div class="active-plan">{{ active_plan.name }}</div>
+
+    <div class="active-plan-info">
+        <div class="active-plan-name">{{ active_plan.name_i18n }}</div>
+        <div class="active-plan-status">
+            {% if active_plan.actions_locked %}
+                <div class="active-plan-locked">{% icon name="lock" %}</div>
+            {% endif %}
```
MWargelin: Did I get the logic right here, i.e. is `active_plan.actions_locked` what was meant by the plan being locked?

tituomin: Actually we haven't implemented the locking yet. `actions_locked` was a good guess but we plan to have some feature which locks everything, not just actions.

admin_site/templates/wagtailadmin/base.html
```diff
@@ -18,7 +18,23 @@
...
+            {% if active_plan.version_name %}
+                <div class="active-plan-version-name">{{ active_plan.version_name }}</div>
+            {% endif %}
+            {% if active_plan.is_live %}
+                <div>
+                    <span class="active-plan-live">live</span>
+                </div>
+            {% endif %}
```
tituomin: This is good, but I realized there is a bug in Plan.is_live. It should compare the date to the current date. But not a problem wrt this PR.

Another thing is that I think if it's not live, we want to display that info as well, we could use the term "draft" for the opposite state. Or "internal" is another option.

admin_site/templates/wagtailadmin/base.html
```diff
@@ -18,7 +18,23 @@
...
+        <div class="active-plan-status">
+            {% if active_plan.actions_locked %}
+                <div class="active-plan-locked">{% icon name="lock" %}</div>
+            {% endif %}
+            {% if active_plan.version_name %}
+                <div class="active-plan-version-name">{{ active_plan.version_name }}</div>
+            {% endif %}
+            {% if active_plan.is_live %}
+                <div>
+                    <span class="active-plan-live">live</span>
```
tituomin: We should localize these UI strings.

admin_site/templates/wagtailadmin/base.html
```diff
@@ -18,7 +18,23 @@
     {% else %}
         <img src="{% static 'images/kausal-avatar-dark-green.svg' %}" alt="Kausal Logo" width="80" />
     {% endif %}
-    <div class="active-plan">{{ active_plan.name }}</div>
+
+    <div class="active-plan-info">
+        <div class="active-plan-name">{{ active_plan.name_i18n }}</div>
+        <div class="active-plan-status">
+            {% if active_plan.actions_locked %}
+                <div class="active-plan-locked">{% icon name="lock" %}</div>
```
tituomin: We probably want an aria-attribute for accessibility, some verbal description of the state in addition to the icon

admin_site/templates/wagtailadmin/base.html
tituomin: It's good to have this for testing though, and we might just comment it out for now, before deploying.

tituomin (changes requested):
I will still test how this looks. Left some requests.

admin_site/templates/wagtailadmin/base.html
MWargelin: I thought it was something like that, good that that was cleared. I commented the code out for now with a TODO message.

admin_site/templates/wagtailadmin/base.html
MWargelin: Good catch with the `is_live` bug. Seemed like a small thing so fixed it here anyway, is this what you meant? 43ee9803182fb97637651f2b8156a16b111c866f

I changed this now to print "draft" if the plan is not live.

admin_site/templates/wagtailadmin/base.html
MWargelin: localised the strings :+1:

admin_site/templates/wagtailadmin/base.html
MWargelin: I'm not very familiar with aria and accessibility in general, I added `aria-label` that says `locked plan` (translated), does that sound right?

MWargelin:
@tituomin I believe I've now implemented all the requested changes, is this good to go now?

PR #300, opened 2024-12-02
Fix problem with mixin base class

tituomin:


bbliem:
I don't foresee any problems with this.

PR #301, opened 2024-12-03
Reuse report Excel export for UI

tituomin:
Reuse our existing backend excel report generation code to dynamically generate a "report" to be used in the Watch UI dashboard export, based on the fields of the ActionListPage.dashboard_columns.

Most of the work is just making sure all the blocks exist for the report that existed for the dashboard export in the UI and that they work in a similar way.


tituomin:
UI PR that uses this https://github.com/kausaltech/kausal-watch-ui/pull/438

Bhavatu (approved):
LGTM. A pretty large PR, so i might have missed something while reading it through. 

But you have some tests, and i trust your own testing.

PR #302, opened 2024-12-09
Fix completed report ordering

tituomin:
For a completed report, the custom ordering of actions was not respected. In practice, it meant that actions were listed in completion order.


MWargelin (approved):
Looks good, tested this and it works :+1:

tituomin:
Thank you for testing!

PR #303, opened 2024-12-11
Support continuous implementation phases also in reports

tituomin:


reports/blocks/action_content.py
```diff
@@ -24,7 +25,9 @@ def get_help_label(self, value: Model):
         return None
 
     def get_help_panel(self, block_value, snapshot):
-        value = self.value_for_action_snapshot(block_value, snapshot) or ''
+        value = self. value_for_action_snapshot(block_value, snapshot) or ''
```
bbliem: A space crept in.

reports/blocks/action_content.py
```diff
@@ -24,7 +25,9 @@ def get_help_label(self, value: Model):
...
+        if isinstance(value, Promise):
+            value = str(value)
```
bbliem: Maybe it doesn't occur for our usage, but did you consider that it could be a promise that doesn't evaluate to a string?

reports/report_formatters.py
```diff
@@ -505,9 +506,15 @@ def get_graphene_value_class_properties(self) -> GrapheneValueClassProperties:
...
-        if implementation_phase_id:
-            return ActionImplementationPhase.objects.get(id=implementation_phase_id)
-        return None
+        if not implementation_phase_id:
+            return None
+        try:
+            implementation_phase = ActionImplementationPhase.objects.get(id=implementation_phase_id)
+        except ActionImplementationPhase.DoesNotExist:
+            return _('Unknown')
+        if snapshot.action_version.field_dict['schedule_continuous'] and implementation_phase.identifier == 'completed':
```
bbliem: Maybe better to use `field_dict.get('schedule_continuous')`?

I'm also not sure if hard-coding a magic identifier value is a good idea, especially if it's not documented anywhere. It would probably be more robust to add something like a Boolean.

reports/report_formatters.py
```diff
@@ -518,7 +525,15 @@ def extract_action_values(
         pk = action.get('implementation_phase_id')
         if pk is None:
             return [None]
-        return [str(report.plan_current_related_objects.implementation_phases.get(int(pk), f"[{_('empty')}]"))]
+        implementation_phase = report.plan_current_related_objects.implementation_phases.get(int(pk))
+        result = None
+        if not implementation_phase:
+            result = f"[{_('empty')}]"
```
bbliem: I think `f"[{_('...')}]"` is a bit too much. One of the rare cases where I'd favor something like `'[%s]' % _('empty')`.

reports/report_formatters.py
```diff
@@ -518,7 +525,15 @@ def extract_action_values(
         pk = action.get('implementation_phase_id')
         if pk is None:
             return [None]
-        return [str(report.plan_current_related_objects.implementation_phases.get(int(pk), f"[{_('empty')}]"))]
+        implementation_phase = report.plan_current_related_objects.implementation_phases.get(int(pk))
+        result = None
```
bbliem: This line can be removed.

reports/report_formatters.py
```diff
@@ -518,7 +525,15 @@ def extract_action_values(
...
+        if not implementation_phase:
+            result = f"[{_('empty')}]"
+        elif action.get('schedule_continuous') and implementation_phase.identifier == 'completed':
+            result = gettext('Continuous Action')
```
bbliem, line 531: Why `gettext` here and not `gettext_lazy` like above (as `_`)?

reports/blocks/action_content.py
tituomin: I ditched the Promise instead. For consistently, lazy strings are never used in the reporting context now. Best restrict lazy translation usage to static class contexts where it's actually needed.

reports/report_formatters.py
tituomin: Fixed.

tituomin: The hard coding is now a fact of life, I have a plan to get rid of it but it is pending on a small refactoring of phases. The plan is to make the _last_ phase always the completed one. I added an abstraction layer which we will use for this in the future. Ie. a method.

PR #305, opened 2024-12-14
Remove duplicate plan menu item

bbliem:
This (and removing the associated modeladmin classes) required some changes to the tests.

### DIFF
--- a/actions/wagtail_admin.py
+++ b/actions/wagtail_admin.py
@@ -5,9 +5,12 @@ import typing
 from functools import cached_property
 
 from django.core.exceptions import ValidationError
+from django.db import transaction
+from django.db.models import Model, ProtectedError
 from django.urls import re_path, reverse
 from django.utils.translation import gettext_lazy as _
 from wagtail.admin.filters import WagtailFilterSet
+from wagtail.admin.messages import validation_error
 from wagtail.admin.panels import (
     FieldPanel,
     InlinePanel,
@@ -17,13 +20,14 @@ from wagtail.admin.panels import (
 from wagtail.admin.ui.tables import BulkActionsCheckboxColumn, Column
 from wagtail.admin.widgets.button import ButtonWithDropdown
 from wagtail.coreutils import capfirst
+from wagtail.permissions import ModelPermissionPolicy
 from wagtail.snippets.models import register_snippet
 from wagtail.snippets.views.snippets import IndexView, SnippetViewSet
 
 from dal import autocomplete
 from django_filters import filters
 from wagtail_modeladmin.helpers import PermissionHelper
-from wagtail_modeladmin.options import ModelAdminMenuItem, modeladmin_register
+from wagtail_modeladmin.options import modeladmin_register
 
 from aplans.context_vars import ctx_instance, ctx_request
 
@@ -36,11 +40,12 @@ from admin_site.models import Client, ClientPlan
 from admin_site.permissions import PlanSpecificSingletonModelSuperuserPermissionPolicy
 from admin_site.viewsets import WatchEditView, WatchViewSet
 from admin_site.wagtail import (
-    ActivePlanEditView,
     AplansAdminModelForm,
     AplansCreateView,
+    AplansEditView,
     AplansModelAdmin,
     CondensedInlinePanel,
+    SuccessUrlEditPageModelAdminMixin,
     insert_model_translation_panels,
 )
 from copying.views import PlanCopyView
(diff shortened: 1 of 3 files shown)


actions/wagtail_admin.py
```diff
@@ -586,6 +567,18 @@ class Meta:
         fields = ['clients__client']
 
 
+class PlanPermissionPolicy(ModelPermissionPolicy):
+    def __init__(self):
+        super().__init__(Plan)
+
+    def user_has_permission_for_instance(self, user: AbstractBaseUser, action: str, instance: Model) -> bool:
```
MWargelin: I think this method doesn't actually get called, at least when I add a line that prints something and then open the `Plans` list view, nothing gets printed. If I add a `user_has_permission` method here that prints something, then the prints appear to the console when I open the Plans listing view.

I once dived into this myself and seems like for most of the time Wagtail doesn't call `user_has_permission_for_instance` when checking permissions, instead it calls `user_has_permission`. The best way I found to work around that was to define a custom `user_has_permission` method in the view that then calls `user_has_permission_for_instance`, like here: https://github.com/kausaltech/kausal-watch-private/blob/main/content/wagtail_admin.py#L48 , although I'm not sure if that works for what you were thinking of...

actions/wagtail_admin.py
```diff
@@ -586,6 +567,18 @@ class Meta:
...
+        if not super().user_has_permission_for_instance(user, action, instance):
+            return False
+        if not user.is_authenticated:
+            return False
+        assert isinstance(user, User)
```
MWargelin: Kind of more general question here, I've seen a lot of `assert`s around our code, but to me raising errors seems like the more natural/pythonic way of doing these kind of checks. Is there some rationale for using asserts?

MWargelin (changes requested):
Looks good, but had a couple of questions :slightly_smiling_face:

actions/wagtail_admin.py
bbliem: I try to use assertions only to express conditions that necessarily hold due to our application logic (or at least should hold unless there is some bug). I usually do this in situations where the code would crash later on anyway if the assertion did not hold. For example, here, if `user` is not a `User` instance, I suppose the code would crash in the next line when we try to call `user.get_active_admin_plan()`. I think using assertions this way is nicer than just waiting for the crash because (a) it expresses my intentions when I wrote the code, and (b) if the assertion does not hold, we get an error message that allow us to find the source of the problem faster.

Assertions should not be used in cases where omitting the assertion could lead to unexpected results (e.g., because the code in fact does not crash if the assertion does not hold), since we can't always rely on the assertions being executed. But I try not to use them in those cases. Let me know if you found a place where I do.

Another reason why I often use assertions -- and probably this is the reason for this particular one -- is to make the type checker shut up because it sometimes can't figure things out that we can.

MWargelin: Alright, good to know, this was a new concept for me :+1:

actions/wagtail_admin.py
MWargelin: This might have got forgotten, any comments for this?

bbliem: You're right, it doesn't seem to get called. It's a bit strange, but the entire permission policy doesn't seem to do much. Indeed, I can remove it and the results are still as they should be, it seems: for example, a plan admin can't edit plans other than their own.

I removed this permission policy.

bbliem:
I removed the permission policy altogether. I'm not entirely sure why, but the permissions still seem to be handled as they should. Do you have any idea why? Just requested another review in case I missed something. If you don't have any objections to my removing the permission policy, let's get this PR merged.

MWargelin:
> I removed the permission policy altogether. I'm not entirely sure why, but the permissions still seem to be handled as they should. Do you have any idea why? Just requested another review in case I missed something. If you don't have any objections to my removing the permission policy, let's get this PR merged.

I'm guessing the index view listing works thanks to using [`PlanFilter`](https://github.com/kausaltech/kausal-watch-private/blob/3f1054fde1329f54818a86d4cd0bb5d84a8b2ded/actions/wagtail_admin.py#L597) and when you try to edit a plan it switches over to modeladmin views and the [modeladmin permission helper kicks in](https://github.com/kausaltech/kausal-watch-private/pull/305/files#diff-9e8d5025390277e22dbbcbd60605f7cc5f52b39806e9b97653532da27c8d35ccR179)?

But yeah the permissions seem to work without the permission policy as well, so maybe we should merge this now. We can work on the permission policy once this is fully migrated to snippets.

MWargelin (approved):

PR #307, opened 2024-12-17
Fix/corrupted collection tree

tituomin:
---
  - https://app.asana.com/0/0/1206138442131829


tituomin:
~Don't review quite yet, I noticed a problem in manual testing.~

tituomin:
No it wasn't a problem after all, it's just that I was using a db where the tree had already been corrupted.

bbliem (approved):

PR #308, opened 2024-12-18
Fix/update cluster related objects in plan copy

bbliem:
This should fix an error where copying a plan with drafts and then publishing a draft in the copy would steal related objects (e.g., contact persons) from the original action. That would happen because the serialized data in the revisions still contains references to instances belonging to the original plan.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 /static/
 /.direnv/
 /.*env*
+!/.env.test
 /.git/
 /.github/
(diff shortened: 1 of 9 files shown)


bbliem:
Interesting. The tests worked on my machine, but not in CI. I'm going to check it out.

tituomin:
Code looks ok. I'll test this locally a bit.

tituomin (approved):
Tested, and works!

PR #312, opened 2025-01-02
Avoid saving Indicator, CategoryType, Action accidentally from a different plan

Bhavatu:
Check that the active plan when saving is the same as when the edit form was loaded.
For action, indicator and categorytype.

Doesn't seem to be a problem anywhere else that I can think of (not really in Action either, but doesn't hurt)

### DIFF
--- a/actions/action_admin.py
+++ b/actions/action_admin.py
@@ -213,7 +213,6 @@ class ActionAdminForm(WagtailAdminModelForm[Action]):
             # Organizations can only have at most one role as a responsible party
 
     def save(self, commit=True):
-
         initial_plan_id = self.initial_plan_id
         # Use initial_plan_id to detect mismatch between the active plan and the initial plan on form load.
         if initial_plan_id and str(initial_plan_id) != str(self.instance.plan.id):
@@ -221,8 +220,8 @@ class ActionAdminForm(WagtailAdminModelForm[Action]):
 
             request = ctx_request.get()
             messages.add_message(request, messages.WARNING,
-                                 _("Active plan was changed during the editing of this action. "
-                                   "Action was saved with the original plan: %s")
+                                 _('While editing this action you have switched to a different plan. '
+                                   'This action was still saved with the original plan "%s".')
                                  % initial_plan.name)
             self.instance.plan = initial_plan
 
--- a/actions/action_admin.py
+++ b/actions/action_admin.py
@@ -53,6 +53,7 @@ from admin_site.wagtail import (
     CondensedInlinePanel,
     CustomizableBuiltInFieldPanel,
     CustomizableBuiltInPlanFilteredFieldPanel,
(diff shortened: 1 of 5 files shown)


tituomin (dismissed):
Suggestion: Change validation error into an informational message, save to the old plan.

actions/action_admin.py
```diff
@@ -210,6 +213,19 @@ def clean(self):
             # Organizations can only have at most one role as a responsible party
 
     def save(self, commit=True):
+
```
tituomin: extra line here?

actions/action_admin.py
```diff
@@ -210,6 +213,19 @@ def clean(self):
...
+        initial_plan_id = self.initial_plan_id
+        # Use initial_plan_id to detect mismatch between the active plan and the initial plan on form load.
+        if initial_plan_id and str(initial_plan_id) != str(self.instance.plan.id):
+            initial_plan = Plan.objects.get(id=initial_plan_id)
+
+            request = ctx_request.get()
+            messages.add_message(request, messages.WARNING,
+                                 _("Active plan was changed during the editing of this action. "
```
tituomin: While editing this action you have switched to a different plan. This action was still saved with the original plan "%s".

actions/category_admin.py
```diff
@@ -253,8 +256,25 @@ def get_form_class(self):
...
         super().__init__(*args, **kwargs)
 
+    def save(self, commit=True):
+        obj = super().save(commit)
+        initial_plan_id = self.initial_plan_id
+        if initial_plan_id and str(initial_plan_id) != str(self.plan.id):
+            initial_plan = Plan.objects.get(pk=initial_plan_id)
+            request = ctx_request.get_admin_request()
+            messages.add_message(request, messages.WARNING,
+                                 _("Active plan was changed during the editing of this category type. "
```
tituomin: Same change here.

indicators/wagtail_admin.py
```diff
@@ -281,9 +288,22 @@ def clean(self):
...
 
     def save(self, commit=True):
+        initial_plan_id = self.initial_plan_id
+        # Use initial_plan_id to detect mismatch between the active plan and the initial plan on form load.
+        if initial_plan_id and str(initial_plan_id) != str(self.plan.id):
+            initial_plan = Plan.objects.get(id=initial_plan_id)
+
+            request = ctx_request.get()
+            messages.add_message(request, messages.WARNING,
+                                 _("Active plan was changed during the editing of this indicator. "
```
tituomin: Same. Actually, you if it's not too much effort, you can also parametrize with the model name (though it's sometimes a bit risky for some languages (?) considering ensuring proper grammar).

tituomin (dismissed):
I wanted to remove the passive from the message to make it even clearer what had happened.

tituomin (approved):

PR #313, opened 2025-01-08
[Pending testing] Remove django-helusers

tituomin:
Replace with internal minimal implementation. To be replaced with Azure Entra soon

### DIFF
--- a/admin_site/backends.py
+++ b/admin_site/backends.py
@@ -24,8 +24,7 @@ class AzureADAuth(AzureADTenantOAuth2):
     def get_user_id(self, details, response):
         """Use oid claim as unique id."""
         oid = response['oid']
-        # Replace the pairwise 'sub' field with the oid to better play along
-        # with helusers.
+        # Replace the pairwise 'sub' field with the oid
         response['sub'] = oid
         return oid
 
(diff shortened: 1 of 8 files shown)


admin_site/backends.py
```diff
@@ -67,3 +66,32 @@ def user_data(self, access_token, *args, **kwargs):
...
     def get_user_id(self, details, response):
         return self.id_token['sub']
+
+
+class TunnistamoOIDCAuth(OpenIdConnectAuth):
+    """
+    Soon-to-be-deprecated authentication backend for Tunnistamo login of the City of Helsinki.
+
+    This will be replaced with the standard Azure Entra ID login
+    (ie. AzureADAuth) very soon.
```
MWargelin, line 76: These kinds of statements always make me worried that it will never happen :sweat_smile: Is there something we should/could do to ensure it does, make an Asana task or something?

MWargelin (changes requested):
Looks good to me, this seems to address all heluser mentions in our code and looks clean, I don't know much about our authentication stuff though and don't know about the functionality itself or how to test it :sweat_smile: If you have tested this and are confident this works I'm fine with it :stuck_out_tongue: 

There's one Ruff complaint and I had one small question though, otherwise I think this is good to go :+1:

admin_site/backends.py
tituomin: Yeah, I think I will actually leave this PR as a plan B and let's try to get the AzureAD version working as plan A.

tituomin:
This is rejected in favor of just removing anything Helsinki-specific after Helsinki has switched to Azure (this week)

PR #316, opened 2025-01-15
Handle the enum value ALL_FOR_AUTHENTICATED

tituomin:
Previously this was left unhandled which caused an error to be raised.


MWargelin (approved):
Looks good to me :+1:

PR #318, opened 2025-01-15
Fix rich-text editor language when editing Wagtail pages

bbliem:
Uses the same idea as the hack we used for editing Action instances; however, that one didn't work when editing pages.

https://app.asana.com/0/1202880927552397/1206120724392386/f

### DIFF
--- a/admin_site/draftail_rich_text_area.py
+++ b/admin_site/draftail_rich_text_area.py
@@ -1,44 +1,57 @@
-import re
+from __future__ import annotations
+
 from copy import deepcopy
 
 from django.utils.translation import gettext
 from wagtail.admin.rich_text import DraftailRichTextArea
+from wagtail.admin.rich_text.editors.draftail import DraftailRichTextAreaAdapter
 
 
 # Workaround until https://github.com/wagtail/wagtail/pull/11075 is merged
+def _hack_rich_text_editor_options(options: dict) -> None:
+    type_to_trans_map = {
+        "LINK": gettext("Link"),
+        "DOCUMENT": gettext("Document"),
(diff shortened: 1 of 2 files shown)


admin_site/draftail_rich_text_area.py
```diff
@@ -1,44 +1,57 @@
...
+from __future__ import annotations
+
 from copy import deepcopy
 
 from django.utils.translation import gettext
 from wagtail.admin.rich_text import DraftailRichTextArea
+from wagtail.admin.rich_text.editors.draftail import DraftailRichTextAreaAdapter
 
 
 # Workaround until https://github.com/wagtail/wagtail/pull/11075 is merged
```
MWargelin, line 10: Getting this merged would obviously be the best solution, but looks like it is not moving without some kicking since it's been silent for over a year now. Does it make sense to make a comment there to move it forward or should we just give up on that?

tituomin: I pinged them now.

MWargelin: They answered and seem to be quite receptive of the idea :slightly_smiling_face:  We might need to be active there now :sweat_smile:

bbliem: Yeah. Still, if this happens, it might take months. Should we merge this PR?

MWargelin (approved):
Looks good :+1:

admin_site/draftail_rich_text_area.py
MWargelin: Yeah, we can merge this meanwhile. The code looks good :+1:

PR #319, opened 2025-01-16
fix resolving contact persons from umbrella plan

Bhavatu:
Fixes umbrella plan action list failing when resolving for contact persons

### DIFF
--- a/actions/schema.py
+++ b/actions/schema.py
@@ -1202,7 +1202,6 @@ class ActionNode(AdminButtonsMixin, AttributesMixin, DjangoNode):
     )
     def resolve_contact_persons(root: Action, info: GQLInfo, show_all_contact_persons: bool):
         plan: Plan = get_plan_from_context(info)
-        assert plan == root.plan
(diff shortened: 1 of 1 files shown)


bbliem (dismissed):
I think the solution is fine in the sense that removing the assertion will fix the issue.

But I'd suggest either rewriting it to `assert plan == root.plan or plan == root.plan.parent` (and dropping the `if`) or removing the assertion completely. (In case you're worried about my using `plan == root.plan.parent` instead of `plan.id == root.plan.parent.id`, note that `Model.__eq__()` only compares the PKs anyway.)

Probably better to remove it since theoretically we could have plans within plans within plans, which would break again.

bbliem:
Ah, okay, on second thought, about the `__eq__` thing: Probably your intention was to avoid a DB query. You're right, that's better in your version. But in any case, perhaps better to delete the whole thing. If it were about authorization, we should handle things differently anyway, i.e., not with an assertion.

PR #320, opened 2025-01-16
Migrate notifications to snippets

MWargelin:


admin_site/viewsets.py
```diff
@@ -33,7 +33,7 @@ class WatchEditView[ModelT: Model, FormT: WagtailAdminModelForm](
     PlanRelatedViewMixin,
     ActivatePermissionHelperPlanContextMixin,
     EditView[ModelT, FormT],
-    SetInstanceMixin,
+    # SetInstanceMixin, # TODO: Is this needed? Causes linting errors right now.
```
tituomin, line 36: Do you mean this:
`notifications/wagtail_admin.py:72: error: Definition of "object" in base class "BaseUpdateView" is incompatible with definition in base class "HasObject"  [misc]`

tituomin: I think we generally want to use the context vars, but not sure if they are actually used here.

MWargelin: Yeah that's the error :+1: 

I find the context vars quite confusing, our own custom solution requiring complex code to upkeep while we should be using the requests and instances/objects made available natively by Wagtail :thinking: For the snippets migrated so far looks like we don't need the context vars (I tested that all current snippets using `WatchEditView` seem to be editable and working without `SetInstanceMixin`) so I commented this out for now, but do you think we should hold on to it?

tituomin: No, I guess this is fine. I'm satisfied if you tested it works without it.

tituomin (approved):

PR #321, opened 2025-01-16
fix contact persons in admin ui actionlist

Bhavatu:
Don't redact contact persons in the admin UI actionlist


tituomin (approved):
Fantastic!

PR #324, opened 2025-01-20
Allow deserializing attributes when choices do not exist

bbliem:
Suppose you have an action draft that has a value for a choice attribute. Then you delete the attribute type. Trying to access the action edit page will fail because the code deserializing the value of the attribute will raise `AttributeTypeChoiceOption.DoesNotExist`.

For now, we print a warning instead and let the code proceed with the choice set to `None`.


MWargelin (approved):
Looks good to me :+1:

PR #326, opened 2025-01-20
Fix references in drafts after deleting action-related instances

bbliem:
When deleting an instance related to a model, such as a contact person, drafts of that action may reference the deleted instance, which causes an error when trying to publish the action. This can happen when a plan admin uses the grid editor to change the published version of an action that has a draft.

Here we remove the reference from the revision content so that, when the draft is published, the related instance is created anew instead of trying (and failing) to change the one that doesn't exist anymore.

This is an ad-hoc fix. It should probably be improved. We might need to do something about models other than `ActionContactPerson` and `ActionResponsibleParty` too...

### DIFF
--- a/actions/models/action.py
+++ b/actions/models/action.py
@@ -29,7 +29,7 @@ from modeltrans.translator import get_i18n_field
 from reversion.models import Version
 from wagtail.admin.panels.base import Panel
 from wagtail.fields import RichTextField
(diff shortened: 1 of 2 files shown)


tituomin (approved):
Good. Maybe create a follow-up task to make this systematic.
For that one, I think we should study if we could use Wagtail's reference tracking for this.

bbliem:
Done: https://app.asana.com/0/1206017643443542/1209201114085246/f

PR #330, opened 2025-01-28
Prepend language code to URL path of page if not primary language

bbliem:
This should fix an issue where Wagtail would output the wrong anchor target when linking to a Wagtail page.

An example is reported in [this Slack thread](https://kausaltech.slack.com/archives/CUXBMSZ6U/p1738059396753229).

Right now, for example, we have the following pages in our database:
- `StaticPage` 1595 (title "Espoon ilmastotavoite")
- `StaticPage` 1912 (title: "Espoo's climate goal")

You can try to get the `url` property like this:
```
print(StaticPage.objects.get(id=1595).url)
print(StaticPage.objects.get(id=1912).url)
```
This prints:
```
https://ilmastovahti.espoo.fi/hiilineutraali-espoo/espoon-ilmastotavoite/
https://ilmastovahti.espoo.fi/carbon-neutral-espoo/espoon-ilmastotavoite/
```
The first URL is correct. The second is wrong. Instead, the path should begin with `/en` -- then the URL works.

This PR tries to fix this. I'm not completely sure though if it might have some undesired side effects.

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -182,7 +182,11 @@ class AplansPage(Page):
         if not plan:
             return super().get_url_parts(request)
 
-        return (plan.site_id, plan.site_url, self.url_path)
+        url_path = self.url_path
+        if self.locale.language_code != plan.primary_language:
+            url_path = f'/{self.locale.language_code}{url_path}'
+
+        return (plan.site_id, plan.site_url, url_path)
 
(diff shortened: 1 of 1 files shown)


Bhavatu (approved):
Do not know about the side effects, but it does what you describe.

PR #332, opened 2025-01-29
Setting to ReportType: Only plan admins can mark action to approved for report

Bhavatu:
https://app.asana.com/0/1206017643443542/1209134802084808
Setting in Report Types
image

Dropdown for plan admin:
image

Dropdown for non admin:
image


reports/models.py
```diff
@@ -54,7 +54,8 @@ class ReportType(PlanRelatedModel):
     plan: models.ForeignKey[Plan, Plan] = models.ForeignKey('actions.Plan', on_delete=models.CASCADE, related_name='report_types')  # pyright: ignore
     name = models.CharField(max_length=100, verbose_name=_('name'))
     fields = StreamField(block_types=ReportFieldBlock(), null=True, blank=True)
-
+    only_plan_admins_can_approve_actions = models.BooleanField(
+        default=False, help_text=_('Only plan admins can approve actions for reports of this type'))
```
MWargelin: Maybe we could use the same wording as in the button to make the connection crystal clear i.e. something like `Only plan admins can mark actions as complete for reports of this type`?

MWargelin (approved):
Code looks good to me, I'm approving but had one splitting hairs comment to consider :stuck_out_tongue:

reports/models.py
Bhavatu: You're absolutely right, that was bad wording on my part. Fixed

PR #335, opened 2025-02-03
Migrate organizations to snippets

MWargelin:
Migrate organizations admin view out of using the `modeladmin` library to use the Wagtail native snippet view sets.

Sorry, this will be pretty painful to review :cry: 

Corresponding extensions PR: https://github.com/kausaltech/kausal-extensions/pull/60 . Make sure you have pulled and checked out to the corresponding kausal-extensions branch, as well as built the extensions code (`cd client/src; npm run build`) when testing locally. You might have to also force refresh when browsing the admin UI pages (`ctrl+F5`) to load the correct new JavaScript code.

If you want to see the index view without the custom React frontend, uncomment the code registering OrganizationViewSet (last line of `orgs/wagtail_admin.py`) and comment registering `ExtendedOrganizationViewSet` in kausal-extensions.


orgs/wagtail_admin.py
```diff
@@ -164,27 +163,20 @@ def save(self, *args, **kwargs):
...
-class OrganizationEditHandler(TabbedInterface):
-    def __init__(self, plan, *args, **kwargs):
-        self.plan = plan
-        super().__init__(*args, **kwargs)
-
-    def clone_kwargs(self):
-        kwargs = super().clone_kwargs()
-        kwargs['plan'] = self.plan
-        return kwargs
+class InvisiblePlanPanel(FieldPanel):
```
MWargelin, line 148: Is this a good way to solve this? :thinking: It feels maybe slightly iffy :sweat_smile:

orgs/wagtail_admin.py
```diff
@@ -380,4 +380,4 @@ def get_index_view_buttons(self, user: User, instance: Organization, plan: Plan)
         return buttons
 
 # Extended version is registered in kausal-extensions
-register_snippet(OrganizationViewSet)
+# register_snippet(OrganizationViewSet)
```
MWargelin: Is it okay to leave the view set unregistered here, meaning that this repo doesn't have organisations in admin UI at all, but having extensions on is mandatory? `SnippetViewSet`s don't allow a view set to be registered for same model twice, so a view set with `Organization` as the model can only be registered once.

orgs/wagtail_admin.py
```diff
@@ -103,7 +104,7 @@ class NodeAdmin(ModelAdmin):
     list_display = ('get_as_listing_header', 'get_parent')
     button_helper_class = NodeButtonHelper
     panels = [
-        FieldPanel('name', heading=_("Name")),
+        TranslatedFieldPanel('name'),
```
MWargelin: Using `TranslatedFieldPanel` actually causes bugs if the Organization has different primary language than the Plan. This happens with Sunnydale but maybe not with other cities? Should we work on `TranslatedFieldPanel` or think of other solutions?

I had a quick go at trying to fix `TranslatedFieldPanel` to take the model's primary language into account, but that was not straightforward and led to other bugs.

orgs/views.py
```diff
@@ -29,13 +29,13 @@
 
 class OrganizationViewMixin:
     def get_form_kwargs(self):
-        kwargs = super().get_form_kwargs()
-        kwargs['parent_choices'] = Organization.objects.editable_by_user(self.request.user)
+        kwargs = super().get_form_kwargs()  # type: ignore[misc]
```
MWargelin, line 32: I simply got tired and wanted to get on with it and ignored the typing issues here. The issue is that `self.object` is defined to different types in the jungle of wagtail inheritance tree and I couldn't find a type definition for it here that wouldn't cause typing errors elsewhere. Any thoughts?

orgs/views.py
```diff
@@ -58,25 +88,42 @@ def form_valid(self, form):
...
 
-class OrganizationDeleteView(OrganizationViewMixin, SetInstanceModelAdminMixin, DeleteView):
+def do_rollback():
+    """
+    Raise a Rollback exception.
+
+    To be caught by the transaction.atomic() context manager to rollback the
+    transaction.
+    """
+    raise Rollback()
```
bbliem, line 108: Why make a function of this instead of just raising `Rollback()` wherever it's used?

orgs/wagtail_admin.py
bbliem: It's not the worst of our sins. I don't have a better idea at the moment.

orgs/wagtail_admin.py
bbliem: Personally I think being able to list / edit organizations in *some* form is pretty crucial, and I don't feel good about claiming to be open source when this is missing. But please discuss this with Juha. I guess he gets to decide this.

bbliem (commented):
Thanks, great work. I don't have complaints. Except that in the organization index view when you click an organization (i.e., click its name, not "edit" in the menu), you get sent to a URL that doesn't exist.

orgs/views.py
MWargelin: This was a Ruff complaint, for some reason it didn't like `Rollback()` being raised directly and suggested separating it to a function :man_shrugging:

MWargelin:
> Thanks, great work. I don't have complaints. Except that in the organization index view when you click an organization (i.e., click its name, not "edit" in the menu), you get sent to a URL that doesn't exist.

Hmm strange, this doesn't happen for me. But I had similar issues at some point during developing this, I had to change some of the JS code in extensions. Are you sure you didn't have some old version of the code in some cache? You might have to build the extensions code, and then press `ctrl+F5` on the index view to force fetching the newest JavaScript code :thinking:

orgs/wagtail_admin.py
MWargelin: We shouldn't justify doing things badly because we have done things badly in the past :sweat_smile: But if you don't have any better ideas, I guess we'll go with this then.

orgs/wagtail_admin.py
MWargelin: Thanks, fixed now here: https://github.com/kausaltech/kausal-watch-private/pull/335/commits/0d26ecc66596c527377cf86d78d1031b4e5187c6

PR #336, opened 2025-02-04
Fix placement of modeladmin action buttons for non-editable instances

bbliem:
In a modeladmin index view, when a row corresponding to an instance that is not editable is displayed, the action buttons appear in front of the text in the row. For editable rows, the action buttons appear underneath the text, which in this case is a link.

This CSS hack tries to fix this by setting the width of the action buttons `<ul>` element to 100 %.

Asana task, containing video illustrating the issue: https://app.asana.com/0/1206017643443542/1208668942052483


MWargelin (approved):
Looks good :+1:

PR #338, opened 2025-02-06
Fix model translation form fields for regional variant languages

Bhavatu:
This fixes the model form translation fields to be shown also for languages like es-US, de-CH etc.

e.g in plan edit view:
image

Translation works as should in UI:
image

Asana:
https://app.asana.com/0/1206017643443542/1209285235193134

### DIFF
--- a/admin_site/wagtail.py
+++ b/admin_site/wagtail.py
@@ -6,7 +6,6 @@ from typing import TYPE_CHECKING, Any, Callable, Protocol, Sequence, cast
 from urllib.parse import urljoin
 
 import reversion
-from django import forms
 from django.conf import settings
 from django.contrib.admin.utils import quote
 from django.contrib.auth import REDIRECT_FIELD_NAME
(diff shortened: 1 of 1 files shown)


admin_site/wagtail.py
```diff
@@ -82,7 +82,7 @@ def insert_model_translation_panels[M: Model](
             continue
 
         for lang_code in plan.other_languages:
-            tf = t_fields.get(lang_code)
+            tf = t_fields.get(lang_code.lower())
```
MWargelin: Maybe the function in kausal commons should be used? https://github.com/kausaltech/kausal-backend-common/blob/main/i18n/helpers.py#L41

This makes it transparent what is the reason for modifying the string and will also get it right if for any weird reason the language codes in `plan.other_languages` have an unexpected separator ('-' vs. '_')

MWargelin (approved):
:+1:

PR #341, opened 2025-02-11
Add flag exclude_from_search to Plan model

bbliem:
This can be used to handle cases where a plan is public but should still not show up in search results, for example when checking "search only across other cities' plans". This could make sense for plans that are for demo purposes, or for plans that are, albeit published and available to anyone who knows the URL, still used somewhat internally by a customer.

[Asana task](https://app.asana.com/0/1206017643443542/1209246928297785)

Please not that at this moment it is not clear whether we actually want to implement this. (Discussion ongoing in Asana.)


Bhavatu:
If this is a burning issue, we could go with this straightforward solution for now I think. Once we get the [Plan visibility](https://github.com/kausaltech/kausal-watch-private/pull/287/files#diff-dc328641b7f3f525a4c4b0c439c98d4a33ab246c6e67bc465050c03c41d4795d) PR ready it will basically also accomplish this.

PR #350, opened 2025-03-13
Fix bug causing changes to category choice fields be lost

bbliem:
There is a [bug report](https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1209625334692492) showing that changes to category choice fields are lost. This should fix it.


MWargelin (approved):
great :+1:

PR #357, opened 2025-03-28
Remove references to extensions from migrations

bbliem:
This again fixes migrations so that things don't blow up when `kausal_watch_extensions` does not exist.


MWargelin (commented):
Hmm, looks like these just reappear once we run `python manage.py makemigrations` the next time. Could there be some more permanent solution? :thinking:

bbliem:
Yeah it's annoying. It would be great to have a permanent solution. Seems difficult, but I haven't given it much thought. Merging this for now.

PR #362, opened 2025-04-15
Update dependencies

juyrjola:


.python-version
```diff
@@ -1 +1 @@
-3.12
+3.13
```
MWargelin, line 1: The previous Python version 3.12 is also written in `pyproject.toml` and `.github/workflows/reviewdog.yml`, should those be changed as well?

MWargelin (changes requested):
I've tried to click around the backend but have not encountered anything strange, although I did have to force refresh with `ctrl+F5` before the admin UI worked. Also the public UI seems to work normally. One question about the Python update, otherwise looks good :+1:

MWargelin (approved):
Looks good to me now :+1:

PR #363, opened 2025-04-22
customizations for contact_persons and responsible_parties panels

Bhavatu:
Add support for BuiltInFieldCustomizations for contact person and responsible parties panels. 
This allows for editing and viewing restrictions.

### DIFF
--- a/actions/action_admin.py
+++ b/actions/action_admin.py
@@ -8,6 +8,7 @@ from typing import Any, Iterable, Unpack, cast
 
 from django.contrib import admin, messages
 from django.contrib.admin.utils import quote
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.urls import URLPattern, path, re_path
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


actions/action_admin.py
```diff
@@ -1181,26 +1184,82 @@ def get_admin_urls_for_registration(self):
...
+        customizations = BuiltInFieldCustomization.objects.filter(
+            plan=plan,
+            content_type=ct,
+            field_name=field_name
+        )
+
+        is_visible = True
+        is_editable = True
+
+        for customization in customizations:
```
tituomin: In the model we have:
`    class Meta:
        constraints = [
            models.UniqueConstraint(
                fields=['plan', 'content_type', 'field_name'],
                name='unique_field_customization_per_plan',
            ),
`
So we should be able to use `get` here unless I'm mistaken.

Bhavatu: You're right. Fixed

tituomin (approved):
Code looks clean. If you've tested it with all the different variations, can be merged!

Bhavatu:
> Code looks clean. If you've tested it with all the different variations, can be merged!

Yep, tested that it works with the roles wanted. Caveat that for now this doesn't allow to loosen restrictions for the ones already in place, only to make access tighter. For instance editing permissions for `contact_persons` are set in [get_roles_editable_in_action_by](https://github.com/kausaltech/kausal-watch-private/blob/e177a241447d56afabffecb608a8058e14cfd6c6/actions/models/action.py#L1259)

PR #367, opened 2025-04-24
Add fields related to indicator dashboard to Indicator model

bbliem:
I made some assumptions that are [up for discussion](https://app.asana.com/1/1201243246741462/project/1209894862159781/task/1210055874523633?focus=true).


MWargelin (approved):
Looks good to me in general!

I was left thinking about some details in how the different fields are ordered in the admin UI, but there's some changes to ordering, help texts etc. of old fields as well, so I decided to create a new task to systematically organise the admin UI according to the design: https://app.asana.com/1/1201243246741462/project/1209894862159781/task/1210094715392230?focus=true

One bigger thing I was thinking was that there is the data category color picker in the design: https://www.figma.com/design/ZUHrZpgy6ofVGRee2X1Lwh/-EXT--Indicator-improvements?node-id=197-1710&t=hFuG50ZfQonlm786-1 But the color picker isn't included in this PR. What do we do with that? But I think this can be merged as is and we can implement the color picker separately then.

bbliem:
You're right. Maybe I just forgot about the colors. It's a bit tricky as we do have the `IndicatorDimension` model to link indicators and dimensions, but if we want to have indicator-specific colors for the categories of the linked dimensions, the data model will get ugly, and writing a good UI for managing that in Wagtail will be difficult. Maybe it would be good enough to let a dimension category have the same color in all indicators? I'll talk to Tero about it.

Let's merge this for now then.

PR #368, opened 2025-04-28
Blocks and choosers for indicator dashboard bar

Bhavatu:


indicators/blocks.py
```diff
@@ -108,6 +128,178 @@ class Meta:
...
 
+class DashboardIndicatorChartBaseBlock(StructBlock):
+    """Base class for dashboard indicator chart blocks with common fields and validation."""
+
+    help_text = CharBlock(
+        required=False,
+        help_text=_('Help text for the field to be shown in the UI')
+    )
+    indicator = IndicatorChooserBlock(
+        help_text=_('Choose indicator for data visualization')
```
bbliem: As we use definite articles in other help texts, maybe better to also do it here ("Choose *the* indicator for data visualization").

indicators/blocks.py
```diff
@@ -108,6 +128,178 @@ class Meta:
...
+        required=True
+    )
+
+    # graphql_fields = DashboardIndicatorChartBaseBlock.graphql_fields + [
+    #     GraphQLString('bar_type'),
+    # ]
+
+    class Meta:
+        icon = 'fontawesome-chart-simple'
+        label = _('Indicator Bar Chart')
```
bbliem: The UI design sketch has a colon after "Indicator" here and in other block labels. I think it's a good idea. Might prevent misreadings or mistranslations -- like "what is an indicator bar"?

indicators/blocks.py
```diff
@@ -108,6 +128,178 @@ class Meta:
...
+        help_text = _('Indicator visualization as an area chart')
+
+
+@register_streamfield_block
+class DashboardIndicatorPieChartBlock(DashboardIndicatorChartBaseBlock):
+    show_percentages = BooleanBlock(
+        default=True,
+        required=False,
+        help_text=_('Show percentages')
+    )
```
bbliem: I didn't see "show percentages" in the design sketch. Where is this from?

Note that Tero added a comment today to the sketch:

> OBS! For the pie chart we need an extra field to select/enter a year we want to visualise!

indicators/blocks.py
```diff
@@ -108,6 +128,178 @@ class Meta:
...
+    help_text = CharBlock(
+        required=False,
+        help_text=_('Help text for the field to be shown in the UI')
+    )
+    indicator = IndicatorChooserBlock(
+        help_text=_('Choose indicator for data visualization')
+    )
+    categories = DimensionChooserBlock(
+        help_text=_('Choose the indicator dimension that will be used for categories in the visualization')
+    )
```
bbliem: Should this field be optional? I don't know now myself. I'll ask it Slack.

bbliem (approved):
Thanks. Good work! Just a few details that you may want to take care of and then merge at your leisure.

bbliem:
Also note, by the way, that Tero wrote in the `#tech` channel that it might be a better idea to call the field for choosing a dimension "dimension" instead of "categories", which was the suggestion in the design sketch.

PR #370, opened 2025-05-08
Fix ordering categories through spreadsheets

MWargelin:
When categories were ordered using the spreadsheet functionality, the
corresponding category pages would not follow the ordering. Use the
pre-existing method to synchronize the pages as well.

Also make some fixes to the ordering logic, as by comparing Pages to
CategoryPages the method did not recognise some pages to be identical
and moved pages needlessly. Use Page.specific to be sure to compare
CategoryPages to CategoryPages.

Additionally, avoid some SQL queries by only saving the page if the
relevant attributes have been changed.


Bhavatu (approved):
Makes sense and looks clean. Did not test it out myself but I trust you did.

PR #371, opened 2025-05-08
Tackle some n+1 in actionlist

Bhavatu:
For bremen this results in 950+ --> 57
For san diego: 975 -> 366

Did not really see improvement in speed though locally, so not sure if this is a good idea.

### DIFF
--- a/actions/models/action.py
+++ b/actions/models/action.py
@@ -1030,16 +1030,20 @@ class Action(
             'preceding', 'dependent'
         ).visible_for_user(user, plan)
 
-    def has_contact_person_from_organization(self, organization, include_suborganizations=True):
+    def has_contact_person_from_organization(
+        self,
+        organization: Organization,
+        include_suborganizations: bool = True,
+    ) -> bool:
         persons = list(self.contact_persons_unordered.all())
 
         for person in persons:
-            if person.organization:
-                if include_suborganizations:
-                    if person.organization.path.startswith(organization.path):
-                        return True
-                elif person.organization.id == organization.id:
-                    return True
+            if person.organization is None:
+                continue
+            if include_suborganizations and person.organization.path.startswith(organization.path):
+                return True
+            if person.organization.id == organization.id:
+                return True
         return False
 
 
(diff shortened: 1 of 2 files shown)


actions/models/action.py
```diff
@@ -1026,18 +1026,21 @@ def get_workflow_progress(self) -> tuple[int, int]:
...
-        if include_suborganizations:
-            filter_kwargs = {'organization__path__startswith': organization.path}
-        else:
-            filter_kwargs = {'organization': organization.path}
-        return self.contact_persons_unordered.filter(**filter_kwargs).exists()
+        return ActionDependencyRelationship.objects.qs.all_for_action(self).select_related(
+            'preceding', 'dependent'
+        ).visible_for_user(user, plan)
+
+    def has_contact_person_from_organization(self, organization, include_suborganizations=True):
```
tituomin: We lost some types here.

actions/models/action.py
```diff
@@ -1026,18 +1026,21 @@ def get_workflow_progress(self) -> tuple[int, int]:
...
+        persons = list(self.contact_persons_unordered.all())
+
+        for person in persons:
+            if person.organization:
+                if include_suborganizations:
```
tituomin: No big deal, but I would maybe reduce some of the nesting here, for example put
```
if person.organization is None:
    continue
```
in the beginning, then also maybe:
```
if include_suborganizations and person.orgnization.path.startswith [...]
```

tituomin: (Though the latter changes the logic a bit, it shouldn't make a difference here I think)

tituomin (approved):
This is great! Good work.
I wonder how it's possible that it doesn't result in a speedup.
Have you isolated the offending GraphQL query and run it with curl for example:
```
time curl [...]
```
Since if you use the UI directly maybe it has some caching in place.

tituomin (approved):

Bhavatu:
> This is great! Good work. I wonder how it's possible that it doesn't result in a speedup. Have you isolated the offending GraphQL query and run it with curl for example:
> 
> ```
> time curl [...]
> ```
> 
> Since if you use the UI directly maybe it has some caching in place.

Yes, graphql query was tested by itself and oddly I didn't get any significant speedup

PR #372, opened 2025-05-13
Force immediate update of Wagtail reference index when copying plans

bbliem:
When a model instance is saved, Wagtail enqueues a task to update the reference index. By default, this task is executed when the current transaction is committed. This may cause problems if the code within the transaction not only saves some model instances but also relies on the reference index being kept up to date before the transaction ends.

This commit defines a decorator that temporarily changes this behavior in such a way that the reference index is updated immediately when a model instance is saved. This decorator is then applied to `copy_plan()` to make sure that the reference index is always up to date.

This fixes a bug when copying plans that left references to the original plan in place: The code that is supposed to update the references did not get the correct references from Wagtail's reference index. As `copy_plan()` is usually called in a transaction, at the time the code for updating references ran, the reference index for the copies was still empty because the transaction had not been committed at that point.

### DIFF
--- a/copying/main.py
+++ b/copying/main.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
-from collections.abc import Generator, Iterable
+import dataclasses
+from collections.abc import Callable, Generator, Iterable
 from contextlib import ExitStack
 from copy import copy as shallow_copy
 from functools import singledispatchmethod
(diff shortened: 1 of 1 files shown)


MWargelin (approved):
Looks good and works according to my local test. Great work :+1:

PR #373, opened 2025-05-14
GraphQL field and annotation for action.has_dependency_relationships

tituomin:
Used in big listings to prevent actually querying the dependency details for each action.


Bhavatu (approved):
Makes sense! there was a mypy complaint that could be looked at.

PR #379, opened 2025-05-22
Copy objects owned by pages when copying a plan

bbliem:
This fixes issues for objects with a parental key to a page (at the moment `CategoryTypePageLevelLayout`) that, when copied, would have contained references to the original plan.

This is because our copying logic was not aware of these page-owned objects due to pages (and subsequently page-owned objects) being copied by Wagtail's own copying logic.

We fix this by making our clone visitor aware of pages and page-owned objects. This allows our update references visitor to then find pages and page-owned objects along with their copies.

There were also some changes to `update_indexed_references()` (formerly called `update_streamfield_references()`) to be made along the way.

### DIFF
--- a/copying/main.py
+++ b/copying/main.py
@@ -13,7 +13,9 @@ import wagtail.signal_handlers
 from django.contrib.contenttypes.fields import GenericForeignKey
 from django.contrib.contenttypes.models import ContentType
 from django.core.files.base import ContentFile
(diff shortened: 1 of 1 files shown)


MWargelin (approved):
I can't say I understand much, but the code looks clean and the issue of a stuck draft page is gone judging by a local test. Great! :+1:

PR #380, opened 2025-05-27
Show message instead of empty list when dimension chooser has no results

bbliem:
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1210263436414122?focus=true


MWargelin (approved):
Nice, clean solution! :+1:

PR #382, opened 2025-06-05
More copying tests

bbliem:
I implemented some tests and fixed some issues along the way.


MWargelin (approved):
Once again I can't say I understand much but the code seems sensible and is readable so thumbs up :+1: 

Although, one of the new tests seems to be failing :thinking:

PR #387, opened 2025-07-24
Fix capitalization of model field block labels

bbliem:
In many languages, upper-case letters may occur in places other than the beginning of a sentence. For a Wagtail streamfield block, unless an explicit label is given, Wagtail's standard behavior is to take the implicit label (given by, e.g., the translated name of a model field) and then apply `capfirst` to it. This transforms some upper-case letters to lower case.

We fix this for the blocks we generate for model fields by always supplying an explicit label.

### Before:

Screenshot From 2025-07-24 13-29-23

### After:

Screenshot From 2025-07-24 13-29-39

## Caveats

Note that now it's important that the capitalization in the translated verbose name is as it should appear here. In the example given in the second screenshot, as indicated with greenish yellow, this is currently perhaps not ideal.

Personally I would probably start these German strings with upper-case letters as it's the beginning of the label -- but since the labels are not sentences and the respective words are not nouns, it's at least not incorrect. But perhaps we could update these translations at some point.

Perhaps it would be good to set up a policy to have the translations for verbose names always start with a capital letter (or, more generally, since this may not be correct for all languages, have the translations in such a way that they can stand alone as a label).

### DIFF
--- a/aplans/field_registry.py
+++ b/aplans/field_registry.py
@@ -4,10 +4,12 @@ import importlib
 import re
 import textwrap
 import typing
+from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any, Literal
 
-from django.db.models import Model
+from django.core.exceptions import FieldDoesNotExist
+from django.db.models import Field, Model
 
 from aplans.utils import underscore_to_camelcase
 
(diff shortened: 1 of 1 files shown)


tituomin (approved):
Looks good! Maybe rebase to main. Just a note, if we wanted to ensure the very first letter is uppercase, we could (in theory) use css for that with the ::first-letter pseudo-element but unsure if that would be easy in practice in Wagtail.

PR #391, opened 2025-08-13
Allow disabling Wagtail sending ping requests to the server

tituomin:
The extra ping requests are very distracting locally when reading the server logs.

See https://docs.wagtail.org/en/latest/reference/settings.html#wagtail-editing-session-ping-interval


tituomin:
Especially when debugging with pdb, the pings break the flow and sometimes the alignment of the terminal output of pdb.

MWargelin (approved):
Good idea, and looks good :+1:

PR #392, opened 2025-08-14
Fix/translation field panel fixes

tituomin:
Fixes the behaviour mostly of the OrganizationForm in the case where the organization.primary_language is different from the plan's primary language.


admin_site/forms.py
```diff
@@ -12,8 +12,11 @@
 
 from kausal_common.i18n.helpers import convert_language_code
 
+from aplans.utils import get_language_from_default_language_field
```
juyrjola: Should be imported from `kausal_common.i18n.helpers` instead (and the method from `aplans.utils` removed).

admin_site/forms.py
```diff
@@ -43,18 +46,53 @@ def clean_username(self):
 class WatchAdminModelForm[ModelT: Model](WagtailAdminModelForm[ModelT]):
     plan: Plan | None = None
 
-    def prune_i18n_fields(self):
-        model: type[ModelT] = self._meta.model
-        i18n_field = get_i18n_field(model)
+    def get_languages_to_show(self, i18n_field: TranslationField, plan: Plan) -> set[str]:
```
juyrjola: Seems like the bulk of this method should live under `kausal_common`, as a base class for `WatchAdminModelForm` (and `PathsAdminModelForm`) to inherit form.

admin_site/panels.py
```diff
@@ -1,15 +1,21 @@
...
 from modeltrans.utils import build_localized_fieldname
 from wagtail.admin.panels import FieldPanel, FieldRowPanel, MultiFieldPanel
 
-from kausal_common.i18n.helpers import convert_language_code
+from kausal_common.i18n.helpers import convert_language_code, get_language_from_default_language_field
 
-if TYPE_CHECKING:
-    from aplans.types import WatchAdminRequest
+
+class PrimaryLanguagePanel(FieldPanel):
```
juyrjola: Shouldn't this also be under `kausal_common`?

admin_site/panels.py
```diff
@@ -28,23 +34,12 @@ def clone_kwargs(self):
...
-
-        def is_shown(self):
-            plan = self.request.get_active_admin_plan()
-            ret = super().is_shown()
-            if not ret:
-                return False
-            return self.panel.language in plan.other_languages
-
 
 class TranslatedFieldMixin:
```
juyrjola: Let's try to do it without mixins, as they are notoriously incompatible with Python type hints.

juyrjola (changes requested):

PR #393, opened 2025-08-27
Fix the default for "allow public site login" feature

tituomin:
Must be aligned with the default for hiding unpublished plans behind authentication.

Please note that there should also be some sanity check and validation to prevent unworking combinations but this hopefully fixes the most urgent issue for new plans right now.


MWargelin (approved):
Great :+1:

PR #407, opened 2025-09-24
Update document and image references in rich text fields

bbliem:
Enables copying of plans where some rich text field references an image or document.

Fixes this issue: https://kausaltech.slack.com/archives/C05RGQS52LA/p1758696334218529

### DIFF
--- a/copying/main.py
+++ b/copying/main.py
@@ -717,7 +717,7 @@ def get_cluster_related_objects(instance: Model) -> Generator[tuple[str, Model]]
             assert isinstance(child_relation.remote_field, ParentalKey)
             field_name = child_relation.get_accessor_name()
             manager = getattr(instance, field_name)
-            for x in manager.all():
+            for x in manager.get_object_list():
                 yield field_name, x
 
 
--- a/copying/main.py
+++ b/copying/main.py
@@ -16,7 +16,7 @@ from django.core.files.base import ContentFile
 from django.db.models import Field, ForeignKey, Manager, ManyToOneRel, Model, Q, signals
 from modelcluster.fields import ParentalKey
 from modelcluster.models import ClusterableModel, get_all_child_relations
-from wagtail.fields import StreamField
+from wagtail.fields import RichTextField, StreamField
 from wagtail.models import Page, Revision, Site
 from wagtail.models.media import Collection
 from wagtail.models.reference_index import ReferenceIndex
@@ -30,7 +30,13 @@ from actions.models.category import Category, CategoryType
 from actions.models.plan import Plan
 from actions.signals import create_notification_settings, create_plan_features
 from content.apps import create_site_general_content
-from copying.utils import get_foreign_keys, get_generic_foreign_keys, temp_disconnect_signal, update_streamfield_block
+from copying.utils import (
+    get_foreign_keys,
+    get_generic_foreign_keys,
+    temp_disconnect_signal,
+    update_rich_text_reference,
+    update_streamfield_block,
+)
 from documentation.models import DocumentationRootPage
 from documents.models import AplansDocument
 from images.models import AplansImage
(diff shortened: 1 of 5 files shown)


copying/utils.py
```diff
@@ -94,3 +98,30 @@ def update_streamfield_block(
...
+def update_rich_text_reference(
+    instance: Model,
+    field_name: str,
+    old_referenced_object: Model,
+    new_referenced_object: Model,
+) -> None:
+    """Update a reference to an image or document in a given rich text field."""
+    assert type(old_referenced_object) is type(new_referenced_object)
+    if isinstance(old_referenced_object, AplansDocument):
+        pattern = r'<a\s+[^>]*linktype="document"[^>]*>'
```
MWargelin, line 112: Looks very good overall! I was still wondering could this fail if there are more than one linked documents? (or images?) Should this regex contain the old id to be sure it doesn't match with a tag that is trying to link to another document? Something like this (I can't remember regex syntax by heart so I asked from Claude what it could look like, hope it's not crap):

```
old_id = old_referenced_object.pk
pattern = rf'<a\s+(?=.*linktype="document")(?=.*id="{old_id}")[^>]*>'
```

And then, how would fixing the possible other link work...?

Maybe the code already does this, I just couldn't quite figure it out by reading it so just asking to be sure :stuck_out_tongue:

copying/main.py
```diff
@@ -546,6 +552,18 @@ def update_reference(self, from_object: Model, to_object: Model, source_field: F
             assert getattr(from_object, source_field.name) is copy
             return False
 
+        if isinstance(source_field, RichTextField):
+            field_name, *content_path_rest = content_path.split('.')
+            assert field_name == source_field.name
+            assert content_path_rest == ['']
```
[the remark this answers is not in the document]
bbliem: Sounds like LLM BS. I believe the cases I encountered were indeed "simple field names", and they did end in a dot. We'll need to revise this of course if it turns out in practice that the assertion fails.

copying/utils.py
```diff
@@ -94,3 +98,30 @@ def update_streamfield_block(
...
+
+    old_id = old_referenced_object.pk
+    new_id = new_referenced_object.pk
+
+    def replace_id_in_html_tag(match: re.Match) -> str:
+        return re.sub(rf'\bid="{old_id}"', f'id="{new_id}"', match.group(0))
+
+    old_value: str = getattr(instance, field_name)
+    new_value = re.sub(pattern, replace_id_in_html_tag, old_value)
+    assert new_value != old_value
```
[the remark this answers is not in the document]
bbliem: I don't know about "overly strict". I think it's better for now to assume that (1) HTML generated by Wagtail matches expected patterns, and (2) we end up in this code path when a reference is indeed there, and (3) we end up there exactly once for each reference.

copying/utils.py
bbliem: I'll check later what happens if there are more than one references in the HTML.

Aber your concern that we should make sure not to replace a tag if the ID doesn't match: I believe this is taken care of in `replace_id_in_html_tag()`. We do get a match for tags with a different ID, but we only replace something in the string if the ID is what we're looking for. Or did I overlook something?

MWargelin: Ok, I did some actual testing now and I think it works with multiple links as well, whether the same document is linked multiple times or there different links to different documents. Sorry about causing confusion :sweat_smile: So I think everything is fine, resolving this and approving the PR.

MWargelin (approved):
Looks good!

bbliem:
Thanks for your comments, @MWargelin. Even though the regex code turned out to work as expected, it was good you mentioned this because it's not totally straightforward and there could be some edge cases that cause trouble. I thought this is a case where we should have a unit test, so I wrote one.

While the new unit test did not find a problem with the regex code, it did reveal what seemed to be a bug in the copying code that would cause some references to not be updated. I'm not sure under what conditions exactly that would happen, but I added another commit to this PR that should fix it.

I'll merge this now.

PR #417, opened 2025-10-21
Fix several N+1 queries in admin & GraphQL resolver and improve login perf

juyrjola:
- **Download and update the user avatar only when is has changed**
- **Sync group permissions only at migrate time**
- **Remove minimum version constraints from pyproject.toml**
- **Fix many cases of N+1 query problems**


aplans/context_vars.py
```diff
@@ -8,17 +8,29 @@
...
+
+
+def get_admin_cache(request: HttpRequest) -> PlanSpecificCache:
+    assert hasattr(request, 'admin_cache')
+    return getattr(request, 'admin_cache')  # noqa: B009
+
+
+def get_watch_cache() -> WatchObjectCache:
+    request = ctx_request.get()
+    return getattr(request, 'watch_cache')  # noqa: B009
```
tituomin, line 36: This sort of thing just to get rid of some type errors seems a bit unnecessary, what if we just `request.admin_cache` and then ignore that type error instead. Plus the return type is still `Any` here.

pyproject.toml
```diff
@@ -288,37 +289,38 @@ kausal-paths-extensions = { index = "kausal"}
...
-    "viztracer>=1.0.4",
-    "watchfiles>=1.1.0",
+    "daphne",
+    "django-debug-toolbar",
+    "ipdb",
+    "ipython",
+    "pdbpp",
+    "pre-commit",
+    "uv",
+    "virtualenvwrapper>=6.1.1",
```
tituomin, line 299: This here is still pinned, I assume on purpose.

actions/perms.py
```diff
@@ -339,12 +334,48 @@ def _sync_plan_admin_groups(user):
         _sync_group_page_perms(root_pages, group)
 
 
-def remove_plan_admin_perms(user):
+def _sync_contact_person_group_permissions() -> None:
+    wagtail_perms = Permission.objects.filter(get_wagtail_contact_person_q())
+
+    for plan in Plan.objects.all():
+        if plan.contact_person_group is None:
+            continue
```
tituomin: Add to query set filter for consistency, as above?

tituomin (approved):
Some minor comments, looking good!

pyproject.toml
juyrjola: Yes, `uv add <package>` seems to always add the minimum version constraint. I guess it's okay...

aplans/context_vars.py
juyrjola: Yeah, maybe we should have one contextvar with a dataclass container for all these request-context things (i.e. `request`, `admin_cache`, `watch_cache`). Leaving that as an exercise for later, though!

PR #419, opened 2025-10-23
Add alert for button saving status

woodwoerk:
## Description
Announce the saving text of wagtail publish/ save buttons to screen readers:

1. Inject a new hidden alert div on document loaded if publish actions (`nav.actions`) are present on the page
2. Add listeners to Wagtail progress buttons (used by create and edit views)
3. When the progress value of those buttons change (e.g. if the button is "Saving"), reflect that in the alert for the screen reader.

https://github.com/user-attachments/assets/43ed1ffe-107f-4863-9faf-6fb109b002aa

## Related issue
[🎟️ Asana](https://app.asana.com/1/1201243246741462/project/1205309956497857/task/1211575003917065?focus=true)

### DIFF
--- a/admin_site/static/admin_site/js/accessibility.js
+++ b/admin_site/static/admin_site/js/accessibility.js
@@ -1,12 +1,15 @@
-(function() {
+(function () {
+  const ACTIONS_STATUS_CLASS = 'actions-status';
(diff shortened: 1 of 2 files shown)


tituomin (approved):
Looks as clean as this can be. Still wondering how we could add automatic tests that this doesn't break in the future but it might be very difficult. I guess only e2e could handle it.

PR #420, opened 2025-10-29
Monkey-patch modelcluster bug and take fix for related actions back into use

bbliem:
## Description
We recently changed `related_actions` of the `Action` model to a `ParentalManyToManyField` in order to fix the bug that related actions were not saved. While this fixed the bug, it made the action list in the frontend behave in an odd way. (The actions were no longer grouped by categories.) Closer investigation showed that this is due to a bug in django-modelcluster. I made a PR: https://github.com/wagtail/django-modelcluster/pull/203

Until that PR to modelcluster is merged and taken into use, we can use the monkey-patch proposed by the current PR.

## Related issue
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1211588710979857?focus=true

## Testing
A unit test has been included in my PR to modelcluster.

### DIFF
--- a/actions/apps.py
+++ b/actions/apps.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
 
 _wagtail_image_chooser_viewset_permission_policy = None
 _wagtail_get_base_snippet_action_menu_items = None
+_create_deferring_forward_many_to_many_manager  = None  # type: ignore[var-annotated]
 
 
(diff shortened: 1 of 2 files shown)


tituomin (approved):
Beautiful!

tituomin:
It occured to me though that especially with monkey patches, maybe a regression test would be nice.

PR #422, opened 2025-11-03
Fix saving of action categories

bbliem:
Since we changed the `categories` field to a `ParentalManyToManyField`, the categories did not get saved when saving the action. This was because we relied on `action.categories.add()` committing the changes right away, which stopped being the case.

This commit fixes that by calling `Action.save()` on the action instance yet again after updating the categories. Perhaps one of our two calls to `save()` is now redundant, but I'm not sure which one, if any.

In the course of that I also pulled the call to `super().save()` out of the loop in `ActionAdminForm.save()`. I don't know what the idea was behind putting it into the loop in the first place. It seems like a bad idea though because one of the things `super().save()` does is add a method `save_m2m` to the form that's meant to be called later on, and I think we should not assume that this method will be exactly the same every time we call `super().save()`.

## Related issue
[Asana task](https://app.asana.com/1/1201243246741462/project/1202960757879747/task/1211804598641491?focus=true)
[Slack discussion](https://kausaltech.slack.com/archives/C057MEVG13Q/p1762152888667459)

---

> [!NOTE]
> Ensure action categories are saved by re-saving the object after category updates and refactor save flow to call super().save() once outside the roles loop.
> 
> - **Admin form save flow (`actions/action_admin.py`)**
>   - Call `super().save(commit)` once, store as `obj`, and reuse it instead of repeatedly saving inside the roles loop.
>   - Use `getattr(obj, relation_name)` when processing role-based related formsets; preserve and reconcile original objects.
>   - Update categories via `obj.set_categories(...)`; when `commit` is true, call `obj.save()` to persist `ParentalManyToMany` changes.
>   - Continue attribute persistence via `attribute_type.on_form_save(...)` with commit-awareness.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8f483e155b7809ca1165776ba8b67362be914864. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/actions/action_admin.py
+++ b/actions/action_admin.py
@@ -234,6 +234,8 @@ class ActionAdminForm(WagtailAdminModelForm[Action]):
         if hasattr(self.instance, 'updated_at'):
             self.instance.updated_at = timezone.now()
 
+        obj: Action = super().save(commit)
+
         for _cls, relation_name, __, ___ in MODELS_WITH_ROLES:
             formsets = {}
             # There is a corresponding formset for a role if and only if we can edit objects of that role.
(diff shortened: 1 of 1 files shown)


bbliem:
Oh and now that the categories are a parental many-to-many field, perhaps we could have moved the block that saves the categories above the line that calls `super().save()` and avoid the second call to `obj.save()`. But I'm not sure about the "models with roles" block at the moment, so I wanted to play it safe.

tituomin (approved):

PR #444, opened 2025-11-29
Optionally copy indicators when copying plans

bbliem:
## Description
Adds a flag to `copy_plan` to copy indicators when copying a plan. (Does not provide access in the UI yet.)

------

### Type of Change

### Testing

### Dependencies

---

> [!NOTE]
> **Medium Risk**
> Touches plan duplication logic and Wagtail reference-index rewriting, plus a state-only model migration for indicator value categories; mistakes could lead to broken references or incorrect copied indicator/dimension relationships.
> 
> **Overview**
> Adds an opt-in `copy_indicators` flag to `copy_plan` (and a `--copy-indicators` management command option) to deep-copy a plan’s indicators and their dimensions instead of sharing them, with guardrails that abort when indicators are shared across plans or are common-indicator instances.
> 
> Extends the cloning and reference-rewrite pipeline to traverse cluster-related objects and update Wagtail reference-index entries for indicator-owned rich text, including updating indicator links’ `id` *and* `uuid` in HTML; adds comprehensive tests/factories/fixtures for the new copying behavior.
> 
> Formalizes `IndicatorValue.categories` as an explicit through model (`IndicatorValueCategory`) using `SeparateDatabaseAndState` (reusing the existing m2m table) and updates the DRF serializer to explicitly expose `categories`; also adjusts `PlanCommonIndicator.plan`’s `related_name` and enables reference indexing for `Indicator`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f75c0f3c1baa953a547aa90e0b87d66f0821aaed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


MWargelin:
I was thinking that it's a bit sad if the feature is not really available, so I added it as a flag to the management command (also to make testing this easier for me). So I think we should add it to the management command and here's the code since I did it already:
```
diff --git a/copying/management/commands/copy_plan.py b/copying/management/commands/copy_plan.py
index 7f8d47bf..63eb661d 100644
--- a/copying/management/commands/copy_plan.py
+++ b/copying/management/commands/copy_plan.py
@@ -46,6 +46,11 @@ class Command(BaseCommand):
             help="Supersede original actions by their copies",
             action='store_true',
         )
+        parser.add_argument(
+            "--copy-indicators",
+            help="Make new copies of the plan's indicators for plan copy, instead of linking to the old indicators in both plans.",
+            action="store_true",
+        )

     def handle(self, *args, **options):
         plan = Plan.objects.get(identifier=options['identifier'])
@@ -59,4 +64,5 @@ class Command(BaseCommand):
                 version_name=options['version_name'],
                 supersede_original_plan=options['supersede_original_plan'],
                 supersede_original_actions=options['supersede_original_actions'],
+                copy_indicators=options['copy_indicators'],
             )
```
Feel free to modify, especially I'm sure the help text could be worded better.

MWargelin:
I found a bit annoying bug, similar to some issues we've had with copying before... If you
1. Create indicators `A-original` and `B-original` for a plan
2. Link to another indicator in an indicator's description field (using the specialised indicator link block), e.g. set `A-original`'s description to `Check out indicator B-original`
3. Copy the plan and its indicators (now it has new indicators `A-new` and `B-new`)

-> now `A-new`'s description is still `Check out indicator B-original` instead of linking to `B-new`

MWargelin (changes requested):
The main functionality seems to work very well :) There's an annoying edge case I found (there's a separate comment for that), hopefully it's not a pain to fix... And I think the feature could be exposed in the management command, more on that on another comment as well.

MWargelin:
@bbliem This one is getting forgotten, what should we do with this?

bbliem:
Thanks for the reminder and sorry for the delay. You're right. I now added a unit test for this, but I haven't found a way to fix it yet. I believe we need to add indicators to Wagtail's reference index, but that alone doesn't seem to solve the problem.

bbliem:
Okay, this should be fixed now. It was tricky because Wagtail indexes references from an indicator to somewhere else not as belonging to the indicator itself but to its parent object, which happens to be an organization. And as indicators are not children of the plan, we can't use the clone structure for updating references. What needed to be done is
- re-enabling indexing of indicators (Timo had apparently switched it off for performance reasons at some point),
- writing some code that updates references in indicators via iterating over the organizations, updating references in them (albeit only in-memory) and their children, and making sure we only save their indicator children,
- handling HTML string substitutions for indicator references.

MWargelin (commented):
Looks good and referencing works now, thanks! 👍 

But one more thing... Do you know if copying indicators has somehow been used in production even though this has not been merged...? I found this bug, and I've seen very similar cases in production as techsupport, and only after writing the bug as [Asana ticket](https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1213079041168059?focus=true) I realised this shouldn't have happened yet since this PR has been unmerged 🤔 (I couldn't reproduce in main with normal copying)

bbliem:
In production, indicators should not have been copied when copying plans as this PR's code has not been deployed. There is probably a different reason for what you observed.

bbliem:
Okay, I pushed some new commits.
- Dimension categories are now copied when copying indicators.
- For this I needed to make the through model explicit.
- I extended the unit test.

Thanks for finding the issue. I verified that now, when copying the plan you referred to in the Asana task with `copy_indicators=True`, changing an indicator value with dimension categories now works and changes only the data in the copy. The visualizations in the public UI look fine.

However the situation in production came to be, the issue you saw doesn't seem to happen with the current code (anymore).

MWargelin (approved):
Yeah, looks and works great now. Thanks for all the effort! 👍

PR #452, opened 2025-12-22
Generalize and change ActionStatusUpdate model

bbliem:
It's pretty much untested at this point.


Bhavatu:
I think this one could be closed, as [this](https://github.com/kausaltech/kausal-watch/pull/454) got already merged.

tituomin:
I think this approach was rejected in favor of the ChangeLogMessage one which was implemented.

PR #464, opened 2026-01-13
Feature/change log messages for pages

tituomin:
## Description
Add GraphQL fields for changeLogMessage for two specific page types:
StaticPage and PlanRootPage.

N/A

## Related issue
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1212626798048865?focus=true

N/A

## Additional Notes
The information is exposed in two ways:
1. show the ChangeLogMessageBlock in the body blocks
2. expose the latest published PageChangeLogMessage in the changeLogMessage field of the mentioned page types.

------

### Type of Change

### Testing
   Enable the feature "change history messages" for a plan. Go to Pages -> find the root page and some static page. For both, add the block ChangeHistoryMessageBlock to the body. For both, edit the page and "publish" -> write the change history message and save.

### Dependencies


pages/models.py
```diff
@@ -1021,7 +1046,6 @@ class PageChangeLogMessage(BaseChangeLogMessage):
 
     i18n = TranslationField(
         fields=['content'],
-        default_language_field='page__plan__primary_language_lowercase',
```
bbliem, line 1024: Was there a problem with this?

tituomin: Yes, I'm actually debugging it a bit further. But basically it results in an ORM query and pages don't actually have a FK to plan. I'm now doing some checks to see how it could be fixed and not just removed. But the current state of the feature is that multilingual content is not really supported.

bbliem: If it's not, perhaps we should remove the `i18n` field altogether for now? Perhaps better than saving strings with the wrong language?

tituomin: Should be resolved with my new commits.

bbliem: Okay, from what I can tell, "resolved" means we remove the `i18n1 field and leave this for later?

tituomin: For posterity, it's still there in the base class and doesn't need customization per subclass.

tituomin:
Changes approved in Slack conversation.

PR #479, opened 2026-01-26
Add option to hide values from old reports from the reporting tab

tituomin:
The default is to show the values of old, completed reports

## Description
Some users do not want to clutter the reporting tab with values from all completed reports in the system, only those from eg. the previous reporting period.

image

## Related issue
-

-

## Additional Notes
-
------

### Type of Change

### Testing
    Go to Reports > pick a report. If the report is not marked as complete, mark it as complete. Then you can go to the action edit form -> choose the Reporting tab. You will see values from the report in the tab (non-editable). Go back to the report, set the field to False and back the the action -> the values from that report are gone.

### Dependencies


MWargelin (approved):
Looks great! :)

PR #481, opened 2026-01-29
Audit logging for REST API

tituomin:
## Description
Enable Wagtail's change history audit logging when using REST API

N/A

## Related issue
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1212571853082290?focus=true

N/A

## Additional Notes

The actual implementation is really simple.

The biggest hurdle is the migration trickery in the following commits:
746edfc3 * Duplicate the base class implementation from (Page|Model)LogEntry
a9251ab6 * Make fields non-nullable as they are on the old base class
fe3862cd * Change our log entry models to inherit directly from baselogentry
1eb27779 * Delete fields pointing to old base class instances
567c2863 * Delete old base class instances
0a27c267 * Transfer data from old base class instances to our model table
36dd18f5 * Temporarily add base class fields to our log models
9bda5345 * Temporarily remove base classes from our custom log entry models
ae3ac7a0 * Stash the multi-table inheritance pointers to custom field
f212f718 * Add temporary fields to support removing multi-table inheritance

You don't have to spend time verifying that migration path works, I take full responsibility and will test it still thoroughly. Still, if you want to glance at the code, the idea is:

1. We need bulk django ORM operations to efficiently create multiple *LogEntry model instances when doing bulk operations in the REST API.
2. Previously, we used multi-table inheritance (PlanScopedModelLogEntry inherited from ModelLogEntry)
3. 1 and 2 are not compatible, Django does not support bulk ops for multi-table inheritance.
4. To get rid of multi-table inheritance, we need to copy the data from the ModelLogEntry table to the PlanScopedModeLogEntry table so we don't lose old data. We need to do that carefully in several steps, first duplicating the pointer foreign key, removing the inheritance (and the foreign key), recreating the base class fields in the modellogentry, copying the data from the existing base class instances, removing the base class instances, tweaking the fields.

------

### Type of Change

### Testing
    N/A
    
### Dependencies

---

> [!NOTE]
> **High Risk**
> High risk due to major schema/migration changes that remove multi-table inheritance and copy/delete historical audit log rows; failures could corrupt or lose audit history and impact admin change-history views.
> 
> **Overview**
> Adds automatic Wagtail audit logging for REST API CRUD (including bulk create/update) by introducing `AuditLoggingBulkModelViewSet` and switching key viewsets (`Action`, `Category`, `Organization`, `Person`, `ActionTask`, `Indicator`) to use it.
> 
> Reworks plan-scoped audit log models to support efficient bulk writes by removing multi-table inheritance from Wagtail log entry models, adding a `BulkActionModelList` adapter, updating Wagtail log action registration, and shipping a multi-step data migration to copy existing log data and drop the old parent-table rows.
> 
> Tightens permission typing/behavior (`WatchObjectPermissions`/`OrganizationPermission`) and expands test coverage to assert log entries are created for single and bulk REST operations; also adjusts the JSON test client to handle `204 No Content` responses.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d69c8b56fcb553de7bf353117f9ac6dc536f4415. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


Bhavatu (approved):
Looks good all in all. Maybe fix some of the ruff and mypy complaints

Bhavatu:
> ### 3. Bug: log_bulk_action uses wrong plan source
> **File:** `audit_logging/models.py` **Line:** 21 **Issue:** The `log_bulk_action` method hardcodes the plan from `kwargs['user'].get_active_admin_plan()` instead of using the plan that was explicitly passed in the `BulkActionModelList` instance.
> 
> **Problem:** This is inconsistent with how `log_action` handles regular instances (line 60), where it calls `instance.get_plans()` first and only falls back to the user's active admin plan if that returns empty. The `BulkActionModelList` class is explicitly designed to carry a `plan` attribute (see `audit_logging/utils.py:20-25`), but `log_bulk_action` ignores it.
> 
> **Current code:**
> 
> ```python
> plan = kwargs['user'].get_active_admin_plan()
> ```
> 
> **Should be:**
> 
> ```python
> plan = instances.plan
> ```
> 
> This could cause log entries to be associated with the wrong plan when a user's active admin plan differs from the plan context of the API request (e.g., when an admin manages multiple plans and makes an API request specifying a different plan identifier).


Take a look at this too just in case

PR #508, opened 2026-02-13
Add plan deactivation feature

bbliem:
## Description

This PR introduces the ability to mark plans as inactive, hiding them from public view and non-superuser admin interfaces. Superusers can filter plans by active status and see visual warnings for inactive plans in the admin interface.

Plans can now be marked as inactive via a new `is_active` boolean field that defaults to `True`. When a plan is set to inactive, it becomes hidden from public APIs and is filtered out of the admin interface for non-superuser staff. Superusers retain full visibility and gain access to an admin filter that allows them to view active, inactive, or all plans.

The admin interface provides clear visual feedback for inactive plans: a warning icon appears in the first column of the plan list, and the entire row is highlighted with a warning background color. These visual indicators help superusers quickly identify inactive plans when viewing all plans or specifically filtering for inactive ones. Non-superusers continue to see only active plans without any awareness of inactive plans, maintaining a clean and focused admin experience.

### Changes

- **Plan model**: Added `is_active` field with default value `True` and help text indicating inactive plans are hidden from public view
- **Permission filtering**: Updated `permission_policy` and `User.get_adminable_plans()` to automatically filter out inactive plans for non-superusers in both GraphQL queries and admin interfaces
- **Admin filtering**: Implemented `IsActiveFilter` that defaults to showing active plans, with options for "Active", "Inactive", and "All" (only visible to superusers)
- **Visual indicators**: Added warning icon column and row highlighting (using reusable `warning_cell.html` template and CSS) to help superusers identify inactive plans at a glance
- **Comprehensive testing**: Added extensive test coverage including GraphQL query filtering, permission checks, admin filter behavior, and UI visibility rules

------

### Type of Change

### Testing

### Dependencies


actions/tests/test_plan_is_active.py
```diff
@@ -0,0 +1,222 @@
...
+        plan = plan_factory(is_active=False)
+        assert plan.is_active is False
+
+    def test_plan_can_be_deactivated(self, plan):
+        """Test that an active plan can be deactivated."""
+        assert plan.is_active is True
+        plan.is_active = False
+        plan.save()
+        plan.refresh_from_db()
+        assert plan.is_active is False
```
tituomin: I started questioning if we really need to to test that Django's BooleanField really works as is expected. But I suppose making a plan hidden is so drastic that `test_plan_is_active_defaults_to_true` does make sense. We don't really want to change the default in any case. But the other tests here -- not sure if they are needed at this level since the tests at the other abstraction layers cover all the tricky cases.

tituomin (approved):
One point of hesitation, otherwise nice!
I do feel it might make sense to test this manually still in testing.

actions/tests/test_plan_is_active.py
bbliem: I agree. Actually I think we can remove `test_plan_is_active_defaults_to_true` as well. Even if that doesn't work as expected, we'll notice, and nothing will be broken unexpectedly.

PR #521, opened 2026-02-23
Normalize many-to-many model pks to avoid uniqueness failures

tituomin:
ActionContactPersons and ActionResponsibleParties sometimes get edited so that their contents get swapped and therefore their pks get swapped. This results in uniqueness failures in the clusterablemodel code since it results in a situation where the same wrapped object (person or organization) suddenly gets a new wrapper pk while the old wrapper instance is still in the db. This is due to how clusterablemodel handles changes in steps.

In practice this mostly occurs when a user changes the contents of the inline fieldpanels in the Action edit form so that the contents of the panels get swapped (the organization from the second item is moved to the first item and vice versa). This still keeps the old pks of the wrapping objects (the ManyToMany relations such as ActionResponsibleParty) intact. But if an ActionResponsibleParty pk wrapping an Organization changes, due to the fact that clusterablemodel changes these one-by-one in a loop, there comes a point where the organization, action combination exists already in the db when trying to insert it into the new place.

## Description
Before publishing an action revision, or saving an action from the form, this code makes sure the pks already existing in the db, corresponding to the organizations / people, are reused even though the role might change.

N/A

## Related issue
https://sentry.kausal.tech/organizations/kausal/issues/14046/events/6a37717366c34c9896ffe93bddbd850a/?project=3&referrer=event-or-group-header

N/A

## Additional Notes
N/A
------

### Type of Change

### Testing
    If you want, you can try to manually go to an action with two different orgs in primary responsible party and collaborator  and swap the organizations around. No crash anymore.

### Dependencies

### DIFF
--- a/actions/models/action.py
+++ b/actions/models/action.py
@@ -618,8 +618,8 @@ class Action(
     def _renormalize_revision_items(
         self,
         revision: Revision[Self],
-        content_key: str,
-        wrapped_attr: str,
+        content_key: Literal['responsible_parties', 'contact_persons'],
+        wrapped_attr: Literal['organization', 'person'],
     ) -> None:
         """
         Renormalize PKs in revision content to avoid unique constraint violations.
@@ -639,7 +639,8 @@ class Action(
         db_items: Iterable[ActionResponsibleParty | ActionContactPerson]
         if content_key == 'responsible_parties':
             db_items = list(ActionResponsibleParty.objects.filter(action=self))
-        else:  # contact_persons
+        else:
+            assert content_key == 'contact_persons'
             db_items = list(ActionContactPerson.objects.filter(action=self))
 
         # Define accessors for model instances (used for db_items)
--- a/actions/models/action.py
+++ b/actions/models/action.py
@@ -72,7 +72,7 @@ from .attributes import AttributeType as AttributeTypeModel, ModelWithAttributes
 from .features import PlanFeatures
 
 if typing.TYPE_CHECKING:
-    from collections.abc import Iterable, Mapping, Sequence
+    from collections.abc import Callable, Iterable, Mapping, Sequence
 
     from django.db.models.expressions import Combinable
     from django.db.models.options import Options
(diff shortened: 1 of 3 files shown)


actions/models/action.py
```diff
@@ -543,8 +615,65 @@ def commit_attributes(self, attributes: dict[str, Any], user):
...
+        revision_items = revision.content.get(content_key, [])
+        if not revision_items:
+            return
+
+        # Get current DB state
+        db_items: Iterable[ActionResponsibleParty | ActionContactPerson]
+        if content_key == 'responsible_parties':
+            db_items = list(ActionResponsibleParty.objects.filter(action=self))
+        else:  # contact_persons
+            db_items = list(ActionContactPerson.objects.filter(action=self))
```
bbliem, line 644: I'd add an assertion here to make sure we are indeed dealing with an ActionContactPerson, just in case we ever use this function elsewhere in the future and forget to update this.

actions/tests/test_admin.py
```diff
@@ -378,3 +378,387 @@ def test_inspect_button_preserves_category_type_parameter(
...
+    """
+    Test that swapping organizations between ActionResponsibleParty instances works correctly.
+
+    When swapping organizations between different roles (e.g., org A moves from primary
+    to collaborator, org B moves from collaborator to primary), the final state has
+    each organization appearing only once, so this should be a valid operation.
+
+    This test currently FAILS due to a bug where Django's ORM saves formsets sequentially,
+    causing a temporary duplicate violation of the unique_together constraint.
+    After the fix, the swap should succeed without errors.
```
bbliem: "Currently" means before the code also changed in this PR? I just ran the test and it succeeds. Maybe update the comment here if this PR fixes it?

actions/tests/test_admin.py
```diff
@@ -378,3 +378,387 @@ def test_inspect_button_preserves_category_type_parameter(
...
+
+    # Verify the organizations have been swapped to the correct roles
+    assert party_with_org_b.role == ActionResponsibleParty.Role.PRIMARY, \
+        f"org_b should now be PRIMARY, but has role {party_with_org_b.role}"
+
+    assert party_with_org_a.role == ActionResponsibleParty.Role.COLLABORATOR, \
+        f"org_a should now be COLLABORATOR, but has role {party_with_org_a.role}"
+
+
+def test_action_responsible_party_swap_on_publish_draft(plan, organization_factory, action_factory):
```
bbliem: Is there a reason for using the factory fixtures here (and elsewhere in the PR)? The tests above use `*Factory.create()`, which I personally try to favor because it allows for better typing, at least in my editor, and perhaps also tools like MyPy.

bbliem (commented):
Thanks, I didn't find obvious issues.

PR #538, opened 2026-03-16
Rename legacy Dataset model in indicators app to ExternalDataset

bbliem:
## Description
We are currently implementing light-weight dataset support for indicators based on the Dataset model in kausal-backend-common. However, the `indicators` app already has an old model called Dataset. It looks like a legacy system from the pre-Wagtail days. To prevent confusion, I believe it's best to rename the old Dataset model to ExternalDataset (because that's what it seems to be about).

I believe this takes care of all references to the old names, but I'm not too familiar with the legacy system.

------

### Type of Change

### Testing

### Dependencies


bbliem:
As discussed with @juyrjola, we'll remove those models instead.

PR #540, opened 2026-03-24
Admin UI for Indicator factor/Computed datasets 

Bhavatu:
## Description
Brief description of the changes made in this PR.

## Related issue
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1213419066714256

Needs these:
https://github.com/kausaltech/kausal-backend-common/pull/85
https://github.com/kausaltech/kausal-extensions/pull/196

## Additional Notes

------

### Type of Change

### Testing

- [x] **New strings are translatable** (all user-facing text uses i18n)

### Dependencies


indicators/wagtail_admin.py
```diff
@@ -383,8 +591,128 @@ def save(self, commit=True):
...
+
+            # Auto-create/update computations for each factor
+            if obj.dataset_schema is not None:
+                self._sync_factor_computations(obj.dataset_schema, metrics_formset)
+
+            # If all factors were removed, clean up the now-empty schema and dataset
+            if obj.dataset_schema is not None and not self._has_new_factors(metrics_formset):
+                self._cleanup_empty_schema(obj)
+
         return obj
```
tituomin: We have tried to use the convention that the bulk of the code should be as de-intended as possible.

So instead of:

```
if condition:
    # lots of indented lines
return obj
```

We could have:

```
if not condition:
    return obj

# lots of lines with 4 chars less indentation
```

indicators/wagtail_admin.py
```diff
@@ -339,8 +489,61 @@ def clean(self):
...
+    def _ensure_dataset_schema(self, indicator: Indicator) -> DatasetSchema:
+        """Auto-create a DatasetSchema for the indicator if it doesn't have one."""
+        if indicator.dataset_schema is not None:
+            return indicator.dataset_schema
+
+        schema = DatasetSchema.objects.create(
+            name=str(_('Indicator factors')),
+            time_resolution=DatasetSchema.TimeResolution.YEARLY,
+        )
+        # Scope the schema to the indicator instance (per indicators.md architecture)
```
tituomin: Where is indicators.md ?

aplans/dataset_config.py
```diff
@@ -27,6 +33,140 @@ def schema_default_scope():
...
+    if indicator is None:
+        return {}
+
+    from kausal_common.datasets.models import DatasetSchema
+
+    time_res = dataset.schema.time_resolution if dataset.schema else None
+
+    def _normalize_date(d: date) -> date:
+        if time_res == DatasetSchema.TimeResolution.YEARLY:
+            return d.replace(month=1, day=1)
```
tituomin, line 165: We have had the convention for yearly indicators that their date is 31.12. YEAR.
The reason has something to do with the fact that it's often the cumulative value for the whole year, so often it kind of makes sense to record the value for the end of the year. There might have been other reasons too. In some places we are assuming this, for example https://github.com/kausaltech/kausal-watch/blob/feature%2Findicator-computations-inline-panel/indicators/api.py#L225-L230

Not sure what kind of problems this could cause, but maybe best be consistent. Could even make a util function to be used everywhere.

tituomin (changes requested):
A few comments.

aplans/dataset_config.py
Bhavatu: The dataset editor uses January 1 (YYYY-01-01) for yearly DataPoints. ` _normalize_date()` normalizes IndicatorValue dates from Dec 31 -> Jan 1, so they match the DataPoint dates in the computation lookup.

indicators/wagtail_admin.py
Bhavatu: Added it to docs/architecture/

tituomin (approved):

PR #545, opened 2026-04-01
Improve organization handling in plan creation form

tituomin:
## Description
Change the way a new Organization is created from within the plan creation form. Previously it was created using the create tab of the chooser widget -- however, it was impossible to guarantee that the new Organization gets the same primary_language as the plan and in fact the Organization didn't get any language set. Now we create the Organization as part of handling the form so the language will always be in sync. 

Before:
localhost_8000_admin_actions_plan_create_before

After:
localhost_8000_admin_actions_plan_create_new

## Related issue
Part 1 of making sure we never end up with Organizations without primary language.

N/A

## Additional Notes
N/A
------

### Type of Change

### Testing
    Go to plans -> create new. Play around.
    
### Dependencies


Bhavatu (approved):
Looks good, although maybe fix the codex point - sounds valid.

PR #558, opened 2026-04-10
Optimize pledge commitment count to avoid N+1 queries

bbliem:
This is something Claude came up with when I fed it an unrelated task. Seems to make sense, but it would be good to have another opinion.

## Description
Annotate the pledges queryset with _commitment_count in resolve_pledges() so the count is fetched in a single query instead of one per pledge. The resolver falls back to .count() for single-pledge queries.

------

### Type of Change

### Testing

### Dependencies


Bhavatu (approved):
Makes sense to me.

PR #570, opened 2026-04-20
Migrate from modeltrans to wagtail-localize in Pledges

tituomin:
## Description

Pledges contain a StreamField which is incompatible with modeltrans. We remove modeltrans completely for pledges and translate everything with the same translation concept as Wagtail Pages.

This makes Pledge.body translatable.

Brief description of the changes made in this PR.

image

## Related issue
Translating site to Spanish

N/A

## Additional Notes
This way of translation is so far used only for Wagtail Pages. But translating a StreamField is not supported by modeltrans.

Synchronising the translations (structural changes to original streamfield contents) is not supported, it needs another PR.

------

### Type of Change

### Testing
    Enable pledges, add more than one plan language, save plan. Add pledge. See the right panel to switch languages.
    
- [X] **Umbrella plan structure** integration implemented

### Dependencies


bbliem:
It's a pretty large and complex PR. If I wanted to give this a thorough review, it would take quite a bit of time. I understand we are under significant time pressure for this one. So I gave this to Claude and Codex for now. Note that the Codex integration also posted some comments to this PR already. I checked out a couple of the issues the LLMs found and they seemed valid. Can you check?

# Claude review

**Overall**: Solid architectural move. The row-per-locale approach using `TranslatableMixin` is the right call for making `body` (StreamField) translatable. The migration chain is well-structured, tests are comprehensive, and the GraphQL/admin changes are thorough. A few issues to flag:

---

## High Severity

### 1. Migration language normalization mismatch

Migration 0170's `_normalize_language_code` does `.replace('_', '-').lower()`, producing e.g. `en-gb`. But the runtime `convert_language_code(code, 'wagtail')` produces `en-GB` (lowercase language, uppercase region). Both the migration and runtime use `iexact` lookups so this won't cause duplicate locales, but locale rows created by the migration will have lowercase codes (`en-gb`) while any locales created at runtime will have mixed-case (`en-GB`). This creates inconsistency in the DB. Consider using the same normalization in the migration.

### 2. `get_locale_for_language_code` has a create side effect

`actions/models/pledge.py:201-207` — This static method silently creates a `Locale` if one doesn't exist. It's called from `_get_active_wagtail_locale` (via `resolve_pledge`), from `Plan.save()` sync logic, and from the mutation flow. If called with a bad or unexpected language code, it would create a junk locale. The `convert_language_code` call will raise `ValueError` for truly invalid codes, but any valid-format code that's not actually configured for the plan would still create a Locale row. Consider separating the "get" from the "get_or_create" use cases.

---

## Medium Severity

### 3. Orphaned translations when languages are removed from a plan

`Plan._sync_pledge_locale_copies_for_plan_languages` only creates missing locale copies — it never removes copies for languages that were dropped from the plan. If a plan goes from `['en', 'fi', 'sv']` to `['en', 'fi']`, the `sv` pledge rows persist and are invisible in the admin (filtered to primary locale) but remain in the DB and could appear in GraphQL queries if someone explicitly sets `@locale(lang: "sv")`. This may be intentional (preserve data) but is worth documenting.

### 4. `_get_active_wagtail_locale` is plan-unaware

`actions/schema.py:217-232` — This function resolves the active locale from Django's `get_language()` thread-local, falling back to `Locale.get_default()`. The default locale may not match the plan's primary language. If a client requests a language that isn't configured for the plan, or no language is active, pledges could be returned in an unexpected locale (or not at all, since no rows exist for that locale). The page-based resolution for Wagtail Pages has the same constraint, so this might be acceptable within the existing patterns.

### 5. `PledgeCommitment` constraint doesn't enforce canonical storage

`actions/models/pledge.py:406` — `unique_together = [('pledge', 'pledge_user')]` allows a commitment against any locale copy. The code correctly routes commits to the primary translation in the mutation, but there's no DB-level enforcement that commitments only attach to primary-locale pledges. If any other code path creates a commitment against a non-primary copy, the cross-locale count queries (`filter(pledge__translation_key=...)`) will still work correctly, but the data model becomes ambiguous.

### 6. N+1 in `_sync_pledge_locale_copies_for_plan_languages`

`actions/models/plan.py:817-833` — For each source pledge, `ensure_locale_copies()` makes several queries (fetch existing translations, fetch action IDs, look up locales, create copies). For a plan with many pledges and many languages, this could be slow. Not urgent if plans typically have few pledges, but worth noting.

---

## Low Severity / Nits

### 7. Migration 0169 contains unrelated `AlterField` operations

Lines 54–262 of migration 0169 are language-choice-list updates for `attributetype`, `categoryicon`, `commoncategoryicon`, `commoncategorytype`, `plan`. These are noise from `makemigrations` picking up other field changes. Ideally they'd be in a separate migration to keep the pledge migration focused. Not a blocker, but makes the migration harder to review and reason about.

### 8. `_i18n_attribute_panels` is now silently discarded

`actions/pledge_admin.py:165` — `_i18n_attribute_panels` is captured and ignored. This means attribute type translations are no longer editable in the pledge admin. Is this intentional? If attribute types have translatable fields, there's no way to edit them from the pledge form anymore.

### 9. `handle_admin_save` only runs on `create`

`actions/models/pledge.py:246-250` — Locale copies are only created when a pledge is first created in the admin. If new languages are added to a plan, copies are synced via `Plan.save()`, which is correct. But if someone creates a pledge via the ORM/shell without going through admin, `handle_admin_save` won't fire. This is probably fine in practice, but worth noting for completeness.

### 10. Redundant formatting changes in `pledge_admin.py`

The diff includes whitespace/style changes (`icon_name = ''` → `icon_name=''`, reformatted multi-line conditionals, rearranged `mark_safe` call) that are unrelated to the feature. These should ideally be in a separate commit to keep the diff reviewable.

---

## What looks good

- **Migration chain** is well-ordered: add fields → migrate data → remove old field. The intermediate `unique_together` updates in 0170 correctly allow locale copies before the final constraint.
- **`copy_for_translation` flow** in `ensure_locale_copies` correctly generates new UUIDs and copies action M2M relationships.
- **Commitment counting** properly spans all locale copies via `translation_key` subqueries, avoiding double-counting.
- **`CommitToPledgeMutation`** correctly routes to the primary translation so commitments are canonical.
- **Plan cloning** (`copying/main.py`) properly remaps `translation_key` UUIDs so cloned plans get distinct translation groups.
- **Test coverage** is strong: locale-aware GraphQL queries, translation resolution, admin filtering, export, commitment sharing, plan language sync.
- **Wagtail hook** to hide the "Translate" button is a nice touch since translations are managed via locale switching.

# Codex review

## Findings

### 1. GraphQL locale filtering can hide all pledges instead of falling back to the plan's primary pledge

`actions/schema.py` resolves a global active Wagtail `Locale`, then the slug/list/action pledge resolvers filter directly on that locale:

- `actions/schema.py:220` (`_get_active_wagtail_locale`)
- `actions/schema.py:697` (`Plan.resolve_pledge` by slug)
- `actions/schema.py:707` (`Plan.resolve_pledges`)
- `actions/schema.py:1788` (`Action.resolve_pledges`)

This is not plan-aware. If the requested or active language has a `Locale` row but this plan has no pledge copy in that locale, the API returns no pledges instead of falling back to the primary-language pledge.

Example: a plan whose primary language is `fi` queried without `@locale` while Django's active/default language is `en` returns an empty pledge list. Similarly, `@locale(lang: "de")` returns empty for an `en`/`fi` plan if a German `Locale` exists globally from another plan.

Modeltrans previously kept one row and field fallback, so this is a behavioral regression. A safer resolver would choose the requested locale only when that locale exists for the pledge/plan, otherwise fall back to `root.primary_language`.

### 2. The Wagtail snippet locale switcher is effectively broken for pledges

`actions/pledge_admin.py:449` (`PledgeViewSet.get_queryset`) always filters to the plan's primary locale. Wagtail's generic `IndexView` then applies its own locale filter from `?locale=...` for `TranslatableMixin` models. So when an admin switches the pledge listing to `fi`, the final query becomes "primary locale AND fi locale", which returns no rows.

This also conflicts with the new hook comment saying pledge translations are managed via locale switching. The primary edit page may still expose translation links in the side panel, but the normal Wagtail locale listing path and locale-specific exports will be empty for non-primary locales.

The viewset should probably filter by plan and annotate shared commitment counts, but let Wagtail's locale filter select the current locale, or explicitly use the selected request locale instead of `plan.primary_language`.

## Verification

I reviewed the latest commit diff and the relevant Wagtail snippet locale/filtering code. I did not run the full test suite.

Suggested follow-up tests:

1. Add GraphQL tests for a non-`en` primary plan and for a requested locale that exists globally but not for that plan.
2. Add an admin queryset test for `?locale=fi` proving translated pledge rows appear in the snippet listing.

PR #575, opened 2026-04-28
Fix action change log message never returning for public API

lilia1891:
## Description
Claude Code assistance.
Fixes the public UI action change history block not being displayed after publishing an action with a change message.

The reason for the bug was that the UI received `null` for `action.changeLogMessage` from the backend, so it fell back to showing only the legacy updated-at line.
The `get_public_change_log_message` method looked up `ActionChangeLogMessage` by the revision ID from `live_revision.content['latest_revision']`. But the `revision` field on `ActionChangeLogMessage` is never set in the current create flow: the create URL only passes `?action=pk`, not `?revision=...`.
As a result, the lookup always raised `DoesNotExist` and returned `None`, causing the backend to return `null` for the latest change history message.

This change aligns actions with the same pattern used by Indicator, Category, and Page: return the most recently created message for the object.

## Related issue
[asana](https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1214253552276819?focus=true)

## Additional Notes
The public UI expects action.changeLogMessage to return the latest message when the feature flag and layout block are enabled.

------

### Type of Change

### Testing
  The admin setup is already configured for this action https://sunnydale.watch.fi.kausal.tech/actions/U1: the change log feature is enabled and the “Latest change history message” block is present in the action details page layout.

### Dependencies

### DIFF
--- a/actions/models/action.py
+++ b/actions/models/action.py
@@ -1329,16 +1329,7 @@ class Action(
         return visible_contact_persons
 
     def get_public_change_log_message(self) -> BaseChangeLogMessage | None:
-        if not self.live_revision:
-            return None
-        # When publishing, Wagtail creates a new revision, but the previous revision's id
-        # is stored in the live revision. Currently the change history message is always
-        # connected to that previous version, pre-publishing
-        previous_revision_to_published_revision = self.live_revision.content['latest_revision']
-        try:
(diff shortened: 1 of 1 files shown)


tituomin:
This feature actually works as designed -- the change log message for plans with moderation is composed when submitting for moderation and published when accepting the changes in moderation. It is attached to revisions by design.

PR #577, opened 2026-05-04
Validate request parameters for report export view

MWargelin:
## Description
At the moment client errors, in particular invalid request parameters crash the backend and are reported to Sentry. Validate the parameters and return suitable HTTP statuses explaining the issues.

This is an attempt to address [a Sentry issue](https://sentry.kausal.tech/organizations/kausal/issues/10732), where a trailing `'` in the given action ids crash the backend. While this doesn't solve the root cause, I think this is how backend should work, and hopefully next time we run into the bug we get more relevant info from Sentry to solve the bug.

\-

## Related issue
 [Sentry issue](https://sentry.kausal.tech/organizations/kausal/issues/10732)

\-

## Additional Notes
\-

------

### Testing
    \-

### Dependencies

### DIFF
--- a/reports/tests/test_views.py
+++ b/reports/tests/test_views.py
@@ -1,6 +1,13 @@
 from __future__ import annotations
 
+from unittest.mock import patch
+
+from django.contrib.auth.models import AnonymousUser
+from django.http import Http404
+
 import pytest
+
+from actions.tests.factories import PlanFactory
 
 
 def test_mark_action_as_complete_view_accepts_as_view_kwargs():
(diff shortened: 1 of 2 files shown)


reports/tests/test_views.py
```diff
@@ -59,3 +66,76 @@ def test_mark_report_as_complete_view_accepts_as_view_kwargs():
         if 'invalid keyword' in str(e):
             pytest.fail(f'as_view() rejected a keyword argument: {e}')
         raise
+
+
+@pytest.fixture
+def export_view_plan():
+    return PlanFactory.create()
```
tituomin: This fixture does not really do anything in addition to the already-registered plan fixture, see conftest.py at the root. Basically you can just remove these lines and replace "export_view_plan" with "plan" down below. (You can leave the PlanFactory uses below if they need to tweak some parameters of the model.

tituomin (changes requested):
A bit of unnecessary code in tests, otherwise good!

reports/tests/test_views.py
MWargelin: Thanks! Fixed that now.

PR #590, opened 2026-05-19
Fix Action revisions left in has_unpublished_changes=True, latest_revision=None state

MWargelin:
## Description
Claude-conjured code.

<hr>

The `latest_revision` FK uses `on_delete=SET_NULL`, so deleting a Revision row leaves `has_unpublished_changes` untouched.

- `destructively_trim_db`: after bulk-deleting Wagtail revisions, walk all `DraftStateMixin` models and reset `has_unpublished_changes=False` where `latest_revision IS NULL`.
- `copying/main.py`: in the revision-deserialization exception branch of `_copy_instance_revision`, also clear `has_unpublished_changes` when clearing `latest_revision`, so the copy doesn't inherit the impossible state.

\-

## Related issue
[Slack](https://kausaltech.slack.com/archives/C07R954EV7D/p1778587230284149)

\-

## Additional Notes
\-

------

### Testing
1. Find a plan that is using moderation
2. Edit an action and click `Save draft`
3. Delete the user (and maybe also corresponding person) that made the draft
4. Run destructively_trim_db excluding the plan: `python manage.py destructively_trim_db --exclude-plan <plan-identifier>`
5. The output should say somewhere in there that an Action's `has_unpublished_changes` has been reset:
```
[...]
Reset has_unpublished_changes on 1 Action instances with no latest_revision.
[...]
```

### Dependencies

### DIFF
--- a/actions/management/commands/destructively_trim_db.py
+++ b/actions/management/commands/destructively_trim_db.py
@@ -13,7 +13,7 @@ from django.db import ProgrammingError, connection, transaction
 from django.db.models import Exists, OuterRef
 from django.db.models.functions import Cast
 from django.db.models.signals import post_delete, post_save
-from reversion.models import Revision as ReversionRevision, Version
+from reversion.models import Revision as ReversionRevision
 from wagtail.models import DraftStateMixin, ModelLogEntry, PageLogEntry, Revision as WagtailRevision
 
 import factory
@@ -28,7 +28,7 @@ from request_log.models import LoggedRequest
 from users.models import User
 
 if TYPE_CHECKING:
-    from django.db.models import Field, Model, QuerySet
+    from django.db.models import Field, Model
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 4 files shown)


actions/management/commands/destructively_trim_db.py
```diff
@@ -208,6 +210,7 @@ def delete_data(
         # Delete Wagtail revisions without users
         _, by_type = WagtailRevision.objects.filter(user__isnull=True).delete()
         self.print_deleted_instances_by_model(by_type)
```
MWargelin: I'm not entirely sure if the root of the problem actually here, that we blindly delete all revisions without users. I'm thinking a userless revision is a valid thing, it could happen if a user makes a draft, then they quit their job before the draft is approved, and their user is deleted from the system resulting in the draft staying up without a user.

I'm thinking the code as it is now might actually delete these kind of valid revisions. So should we instead here figure out how to exclude revisions that are associated with `plans_to_delete`...?

bbliem: It's possible that userless revisions are valid. I'm not entirely sure if it can be caused by people being removed from the system. I haven't checked now, but in those cases we might keep the `User` (perhaps additionally disabling somehow) but only delete the `Person`.

By the way, a few lines above, we delete personless users, so what I just said may not be too relevant. But this potential issue could perhaps be solved by deleting userless revisions before deleting personless users.

It would indeed be smarter to figure out which revisions to delete via `plans_to_delete`, so if you can think of a good way, feel free to improve that. But perhaps that's difficult. If it turns out to be quite difficult and we can't find a valid case where userless revisions pose a problem for us, we can leave it as it is for now.

copying/main.py
```diff
@@ -1005,7 +1005,11 @@ def _copy_instance_revision(
             f'Failed to deserialize revision for {model_name} pk={instance.pk}, skipping ({type(exc).__name__}: {exc})'
         )
         instance.latest_revision = None
-        instance.save(update_fields=['latest_revision'])
+        if isinstance(instance, DraftStateMixin):
+            instance.has_unpublished_changes = False
+            instance.save(update_fields=['latest_revision', 'has_unpublished_changes'])
```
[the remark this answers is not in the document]
bbliem: Seems like a reasonable comment, no?

bbliem (changes requested):

bbliem:
See https://github.com/kausaltech/kausal-watch/pull/595 for fixes for the two mentioned issues. I'd be happy if you could review my changes, @MWargelin and then merge everything.

In the meantime, since we're in a rush, I'll use my branch for trimming the DB for our upcoming cluster deployment.

PR #592, opened 2026-05-20
Skip copy of collection items whose source file is missing

bbliem:
When the source file is missing on storage, `copy_collection_with_contents` would still call `visit_tree` and save a new row pointing at the source's path, then catch the `FileNotFoundError` from `file.read()` and continue. The result was a new `AplansImage` row sharing its file path with the source — a duplicate-path "time bomb" that orphans both rows the next time either is deleted via standard admin (`post_delete_file_cleanup` deletes the underlying file).

Check `storage.exists` upfront and skip the row entirely. Keep the `FileNotFoundError` catch for the race where the file disappears between the check and the read; in that case delete the just-created row.

------

### Type of Change

### Testing

### Dependencies

### DIFF
--- a/copying/main.py
+++ b/copying/main.py
@@ -24,6 +24,7 @@ from wagtail.models.i18n import Locale
 from wagtail.models.media import Collection
 from wagtail.models.reference_index import ReferenceIndex
 
+import sentry_sdk
 from loguru import logger
 from relations_iterator import AbstractVisitor, ConfigurableRelationTree, RelationTreeIterator  # type: ignore
 
@@ -38,6 +39,8 @@ from content.apps import create_site_general_content
 from copying.utils import (
     get_foreign_keys,
     get_generic_foreign_keys,
+    remove_reference_from_streamfield_block,
+    remove_rich_text_reference_from_field,
     temp_disconnect_signal,
     update_rich_text_reference_in_field,
     update_streamfield_block,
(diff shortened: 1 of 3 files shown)


copying/main.py
```diff
@@ -1050,16 +1050,31 @@ def copy_collection_with_contents(collection: Collection, clone_visitor: CloneVi
...
-        visit_tree(image_or_document, {}, clone_visitor)
-        image_or_document.collection = collection
         file = image_or_document.file
         assert file.name is not None
+        # If the source file is missing on storage, don't create a copy row.
+        # Otherwise the new row would point at the (still-missing) source path,
+        # which means source and copy share an S3 key — a duplicate-path
+        # time bomb that would orphan both rows the next time either is deleted.
+        if not file.storage.exists(file.name):
+            logger.warning(
```
tituomin, line 1239: Frankly not sure if anybody will ever look at these warnings. Is this important enough to send to Sentry?

tituomin (commented):
One question, otherwise good!

tituomin (approved):
Super

copying/main.py
bbliem: Hmm normally I suppose the files should exist on storage, so this may be a sign that something's wrong. I'd suggest to keep this, and if it turns out to be annoying, remove it.

PR #596, opened 2026-05-21
Prevent AplansImageForm from deleting just-uploaded file

bbliem:
When a user re-uploads an image with the same filename via the image edit view, Wagtail's `BaseImageForm.save` unconditionally calls `self.original_file.storage.delete(self.original_file.name)` after the new file is committed to storage. On S3 with `file_overwrite=True` (the django-storages default), `get_available_name` reuses the original path, so the new upload overwrites the old object at the same key — and the subsequent delete then removes that just-uploaded object, leaving the `AplansImage` row pointing at a path that no longer exists on S3.

When a replacement file is being uploaded, wrap
`self.original_file.storage` with an `_OldFileDeleteGuard` proxy for the duration of `super().save()`, so any `delete(name)` whose target matches the form instance's current file path is suppressed. Deletes targeting a different path (the normal case where the new upload lands at a fresh path) still pass through, preserving Wagtail's orphan cleanup. The guard is only engaged when `file` is in `changed_data`, so title-only edits and fresh image creates go through the parent unchanged. Unit tests cover the guard's behavior in isolation, and integration tests exercise the `file_overwrite=True` scenario, the rename-to-a-new-path scenario, and creating a fresh image.

------

### Type of Change

### Testing

### Dependencies

### DIFF
--- a/images/forms.py
+++ b/images/forms.py
@@ -1,24 +1,33 @@
 from __future__ import annotations
 
+from contextlib import contextmanager
 from typing import TYPE_CHECKING, Any
 
 from wagtail.images.forms import BaseImageForm
 from wagtail.log_actions import log
 
 if TYPE_CHECKING:
+    from collections.abc import Iterator
+
     from django.core.files.storage import Storage
+    from django.db.models.fields.files import FieldFile
 
     from images.models import AplansImage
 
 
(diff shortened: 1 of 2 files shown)


tituomin (approved):
Greatness!

PR #626, opened 2026-06-23
Add passwordless email auth for PublicUser

Bhavatu:
## Description
Adds the email-PIN sign-up / sign-in flow on top of the existing anon `PublicUser` model.

### New GraphQL mutations

All under the `pledge` namespace:

- **`signUp(email, termsAccepted, marketingAccepted, anonUuid?)`**. Creates a email-bound signup attempt, (which later upon Pin verification creates an email-bound `PublicUser`), sends PIN
- **`signIn(email, anonUuid?)`**. Sends PIN to an existing account
- **`verifyPin(email, pin)`**. Validates PIN, rotates `user_token`, merges any anon-session pledges, returns the bearer token + current pledge IDs. In case of Signup verification, `PublicUser` is created.

### Anonymous → verified merge

- Both `signUp` and `signIn` accept an optional `anonUuid` (the caller's current localStorage UUID). It's stashed on the `PublicUserSignInAttempt` so the merge intent survives even when
the user verifies on a different device than they signed in from.
- On successful `verifyPin`, the anonymous row's pledge commitments are reassigned to the verified user, deduped on `(pledge, user)`, and the anonymous row is deleted.
- **Cross-email safety:** if `anonUuid` points to a row that has an `email` set (i.e., another verified user), the merge is refused with a Sentry warning so we can't absorb another user's
data via UUID knowledge.

### Other

- New `PublicUserSignInAttempt` model (hashed PIN, expiry, attempts, anon-uuid merge hint).
- New fields on `PublicUser`: `email`, `terms_accepted_at`, `marketing_consented_at`, `email_verified_at`; email is lowercase-normalized on save.
- PIN email is plan-themed. Uses each plan's `BaseTemplate` for branding via a new MJML template.
- All resolver errors tagged with stable `extensions.code` (UI matches on those rather than message strings).

### Backward compatibility

Entirely additive. Existing anon flow (`registerUser` / `commitToPledge` / `setUserData` / `publicUser(uuid:)`) is unchanged; current UI keeps working unmodified after merge.

## Related issue
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1215599133536887
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1215585029682109
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1215589465163003
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1215599134475047
https://app.asana.com/1/1201243246741462/project/1206017643443542/task/1215591653667640

## Additional Notes

------

### Type of Change

### Testing
- [x] **Authorization is tested** (permissions and access controls verified)

- [x] **New strings are translatable** (all user-facing text uses i18n)

### Dependencies


actions/schema.py
```diff
@@ -2437,20 +2467,22 @@ def resolve_category(_root, info: GQLInfo, plan: str, category_type: str, extern
...
+    used so the PublicUser credential never collides with the platform's
+    OAuth/OIDC bearer-token middleware.
     """
     headers = info.context.get_request_headers()
-    auth_header = headers.get('authorization') or headers.get('Authorization')
-    if not auth_header:
-        return None
-    prefix, _, token = auth_header.partition(' ')
-    if prefix.lower() != 'bearer' or not token:
+    token = headers.get(PUBLIC_USER_TOKEN_HEADER.lower()) or headers.get(PUBLIC_USER_TOKEN_HEADER)
```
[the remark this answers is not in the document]
Bhavatu: This has not yet actually been used, and as comment states we need to use a separate header (not Authorization: Bearer) so the PublicUser credential doesnt collide with the platform's OAuth/OIDC bearer-token middleware, which is meant for django users

aplans/schema_context.py
```diff
@@ -324,13 +324,16 @@ def get_cache_key_parts(self) -> list[str] | None:
...
+            self.set_reason('public-user token present')
             return None
 
-        if 'publicUser' in (self.execution_context.query or ''):
-            self.set_reason('publicUser query is identity-sensitive and not cached')
+        query = self.execution_context.query or ''
+        if any(
+            name in query
+            for name in ('publicUser', 'signIn', 'signUp', 'verifyPin', 'registerUser', 'commitToPledge', 'setUserData')
+        ):
```
tituomin: This is a bit fragile, and has some unnecessary parts. First, it's doing substring matching on the unparsed GQL query which could cause false positives.

Second, checking for mutation names is not necessary, since mutations do not use the cache anyway. Here's some LLM slop for your pleasure:

> Two things stand out:

1. The cache only ever runs for QUERY operations — get_cache_key() bails at extensions.py:224 (operation_type != QUERY). Every name in that list except publicUser is a mutation (signIn, signUp, verifyPin, registerUser, commitToPledge, setUserData — all Graphene Mutation.Fields). So those six are already never cached. The substring scan is effectively dead weight for them; only publicUser (a query field) is doing real work.
2. Substring matching on raw query text is the fragile part. name in query will fire on a field alias (assignInfo: ...), a variable name ($signInput), a string-literal argument, or a fragment — false positives that silently disable caching, and the list sits far from where the fields are defined so it rots.

More robust options

Option A — match against selected root fields from the parsed AST (recommended). The document is already parsed in this file (get_first_operation(doc) is used in on_execute just above). Instead of scanning text, pull the operation's top-level selection set and compare field names against a registry:

NON_CACHEABLE_QUERY_FIELDS = frozenset({'publicUser'})

op = get_first_operation(self.execution_context.graphql_document)
selected = {
    sel.name.value
    for sel in op.selection_set.selections
    if isinstance(sel, FieldNode)
}
if selected & NON_CACHEABLE_QUERY_FIELDS:
    self.set_reason('non-cacheable field selected')
    return None

This eliminates the false-positive class entirely (it matches real selected fields, not text), and you can drop the six mutation names since they can never reach this code.

actions/public_user_auth.py
```diff
@@ -0,0 +1,234 @@
...
+    if not email:
+        raise ValueError('email is required; cannot send PIN.')
+
+    minutes = int(PIN_TTL.total_seconds() // 60)
+    base_template = getattr(plan, 'notification_base_template', None) if plan else None
+
+    if plan is not None:
+        plan_name = plan.name_i18n
+        subject = _('Your sign-in code for the %(plan_name)s') % {'plan_name': plan_name}
+        body_intro = _('Here is your sign-in code for the %(plan_name)s:') % {'plan_name': plan_name}
```
tituomin, line 212: Remove the "the"? It makes sense in English for plan names ending in "plan" but not universally, I think.

aplans/graphene_views.py
```diff
@@ -212,13 +212,31 @@ def __init__(self, *args, **kwargs):
...
     def get_cache_key(self, request: WatchAPIRequest, data, query, variables):
         plan_identifier = request.headers.get(PLAN_IDENTIFIER_HEADER)
         plan_domain = request.headers.get(PLAN_DOMAIN_HEADER)
         if not plan_identifier and not plan_domain:
             logger.info('Skipping cache; required HTTP headers missing')
             return None
 
+        if any(name in query for name in self.PUBLIC_USER_AUTH_OPERATIONS):
+            return None
+        if request.headers.get('X-Public-User-Token'):
```
tituomin: Slight mismatch: here uppercase is used, a bit below both upper and lower are fine. Maybe extract to CONSTANT or helper function.

actions/public_user_auth.py
```diff
@@ -0,0 +1,234 @@
...
+
+    plain_body = '{intro}\n\n    {pin}\n\n{ttl}\n\n{ignore}\n\n—\n{powered_by}\nkausal.tech'.format(
+        intro=body_intro,
+        pin=raw_pin,
+        ttl=_('Enter this code when prompted. It expires in %(minutes)d minutes.') % {'minutes': minutes},
+        ignore=_("If you didn't request this code, you can ignore this email."),
+        powered_by=_('Powered by Kausal Watch'),
+    )
+
+    from_email = formataddr(('Kausal', settings.DEFAULT_FROM_EMAIL))
```
tituomin, line 225: Since this is targeting end users would it be less confusing to use the plan name here (still would be kausal.tech email)? They probably don't recognize Kausal. I realize that same problem is with the email address itself, but at least the immediately visible sender name would be more understandable.

notifications/mjml-templates/public_user_pin.mjml
```diff
@@ -0,0 +1,33 @@
...
+                {% trans %}If you didn't request this code, you can ignore this email.{% endtrans %}
+            </p>
+        </mj-text>
+    </mj-column>
+</mj-section>
+
+<mj-section>
+    <mj-column>
+        <mj-text align="center" font-size="12px" color="#666666">
+            {% trans %}Powered by Kausal Watch{% endtrans %}<br>
```
tituomin, line 28: -> "Powered by Kausal"
(I think in the long run we want to start moving into a single unified platform)

actions/public_user_auth.py
```diff
@@ -0,0 +1,234 @@
...
+            verified_user.user_data = merged_user_data
+            verified_user.save(update_fields=['user_data'])
+
+    anon.delete()
+
+    for plan in Plan.objects.filter(id__in=affected_plan_ids):
+        plan.invalidate_cache()
+
+
+def send_pin_email(email: str, raw_pin: str, plan: Plan | None = None) -> None:
```
tituomin: email -> to_email for consistency ?
It could be confused for the email body or something.

actions/models/pledge.py
```diff
@@ -362,6 +365,38 @@ def hash_user_token(raw_token: str) -> str:
...
+    SECRET_KEY; rotating SECRET_KEY invalidates all outstanding PINs,
+    which is fine - users just request a new one (10-minute TTL).
+    """
+    return hashlib.sha256(b'kausal-watch:public_user_pin_pepper:' + settings.SECRET_KEY.encode()).digest()
+
+
+def hash_pin(raw_pin: str, salt: str) -> str:
+    return hmac.new(_get_pin_pepper(), (salt + raw_pin).encode(), hashlib.sha256).hexdigest()
+
+
```
tituomin: I would move authentication-related code and helpers into separate module.

actions/models/pledge.py
```diff
@@ -419,6 +482,168 @@ def regenerate_user_token(self) -> str:
         return raw_token
 
 
+@reversion.register(exclude=['pin_hash', 'pin_salt'])
```
tituomin: Maybe even a separate models/public_user_signin.py or similar for all these. They are concerned with a very specific requirement.

actions/schema.py
```diff
@@ -111,8 +113,19 @@
     IndicatorChangeLogMessage,
     PledgeCommitment,
     PublicUser,
+    PublicUserSignInAttempt,
```
tituomin, line 117: actions/schema.py is getting kind of huge. I think we could have made a completely separate django app (directory in project root) for pledges -- but maybe it's too late for that. Maybe stil separate at least the auth-specific things into a separate file?

actions/schema.py
```diff
@@ -1304,10 +1317,31 @@ class Meta:
...
             'id',
             'uuid',
+            'email',
             'user_data',
+            'email_verified_at',
             'created_at',
         ]
 
+    @staticmethod
+    def _bearer_matches(root: PublicUser, info: GQLInfo) -> bool:
```
tituomin: function name still has "bearer" which might be confused for Authorization: Bearer

actions/schema.py
```diff
@@ -1304,10 +1317,31 @@ class Meta:
...
+
+    @staticmethod
+    def resolve_email_verified_at(root: PublicUser, info: GQLInfo):
+        if not PublicUserNode._bearer_matches(root, info):
+            return None
+        return root.email_verified_at
+
     @staticmethod
     def resolve_commitments(root: PublicUser, info: GQLInfo, plan: str | None = None):
         qs = root.commitments.filter(
```
tituomin, line 1346: Was there a reason the same _bearer_matches check is not used here?

actions/models/pledge.py
```diff
@@ -362,6 +365,38 @@ def hash_user_token(raw_token: str) -> str:
...
+
+    The 6-digit PIN keyspace is small (10**6), so an attacker with DB-only
+    read access could otherwise brute-force the salted SHA-256 in
+    milliseconds and submit the recovered PIN via the public API before
+    the 10-minute TTL expires, minting a bearer token. Peppering raises
+    the bar from "has DB read" to "has DB read AND SECRET_KEY", matching
+    the threat model already paid for by hash_user_token. The label
+    scopes the pepper so it isn't equivalent to keys used elsewhere with
+    SECRET_KEY; rotating SECRET_KEY invalidates all outstanding PINs,
+    which is fine - users just request a new one (10-minute TTL).
```
tituomin: Less verbose and annoying comment would be better. Less "normative": "which is fine" and just summarize essential facts.

actions/models/pledge.py
```diff
@@ -390,6 +425,28 @@ class PublicUser(models.Model):
         verbose_name=_('user token'),
         help_text=_('Opaque secret used as a bearer credential after the user signs up.'),
     )
+    email = models.EmailField(
```
tituomin: Perhaps encrypt.

actions/models/pledge.py
```diff
@@ -419,6 +482,168 @@ def regenerate_user_token(self) -> str:
...
+        defaults = {
+            'pin_hash': hash_pin(raw_pin, salt),
+            'pin_salt': salt,
+            'issued_at': now,
+            'expires_at': now + PIN_TTL,
+            'attempts': 0,
+            'anon_uuid': anon_uuid,
+            'pending_terms_accepted_at': terms_accepted_at,
+            'pending_marketing_consented_at': marketing_consented_at,
+        }
```
tituomin: Some duplication, extract into helper function _get_defaults or similar. Just inject the differing keys+values into the same base defaults dict shared between the two functions

actions/models/pledge.py
```diff
@@ -419,6 +482,168 @@ def regenerate_user_token(self) -> str:
...
+        Validate a raw PIN against the stored hash.
+
+        Increments the attempt counter atomically on every call so parallel
+        verify calls can't overwrite each other's increment and exceed the
+        PIN_MAX_ATTEMPTS ceiling. Returns False if the attempt is expired,
+        the post-increment counter exceeds the cap (a race past the
+        per-mutation pre-check), or the PIN doesn't match. The cap check is
+        strict-greater-than so the Nth attempt - where N == PIN_MAX_ATTEMPTS
+        - is still usable; the per-mutation pre-check via attempts_exhausted
+        catches the (N+1)th request before it ever reaches this code.
```
tituomin: Make less verbose.

actions/perms.py
```diff
@@ -316,6 +317,7 @@ def remove_contact_person_perms(user: UserModel, model: type[Action | Indicator]
     (Pledge, ALL_PERMS),
     (PledgeCommitment, ('view',)),
     (PublicUser, ('view',)),
+    (PublicUserSignInAttempt, ('view',)),
```
tituomin: What is this used for?

actions/schema.py
Bhavatu: This is to be also accessed by anonymous users with no bearer. (Notice the test that makes sure that token users are actually required to use token)

actions/public_user_auth.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+def merge_anon_into_verified(verified_user: PublicUser, anon_uuid: UUID) -> None:
+    """
+    Move pledge commitments from the anonymous PublicUser at anon_uuid into verified_user.
+
+    No-op when the anon_uuid is stale (no matching row), when the row exists
+    but already has an email or user_token set (any past-verification row is
+    off-limits to UUID-based merges, matching _resolve_public_user's rule
+    that any token-bearing row requires the bearer header). Duplicate
```
tituomin: stale

actions/public_user_auth.py
```diff
@@ -0,0 +1,234 @@
...
+        return
+
+    affected_plan_ids = set(
+        PledgeCommitment.objects.filter(public_user=anon).values_list('pledge__plan_id', flat=True).distinct()
+    )
+    existing_pledge_ids = set(PledgeCommitment.objects.filter(public_user=verified_user).values_list('pledge_id', flat=True))
+    PledgeCommitment.objects.filter(public_user=anon).exclude(pledge_id__in=existing_pledge_ids).update(public_user=verified_user)
+
+    if anon.user_data:
+        merged_user_data = {**verified_user.user_data, **anon.user_data}
```
tituomin, line 171: Anon zip code always overrides verified user zip code. But might the latter be more recent, for example intentionally fixing older typo?

tituomin:
Many mentions about bearer tokens in the code comments, even though Bearer is not used anymore. (Though I guess bearer could refer to the concept , not the exact header syntax.

actions/tests/test_graphql_pledge.py
```diff
@@ -1099,13 +1108,760 @@ def test_set_user_data_with_bearer_token_authenticates(self, graphql_client_quer
...
+        assert 'errors' in response
+        assert 'Invalid PIN' in response['errors'][0]['message']
+        assert response['errors'][0]['extensions']['code'] == 'INVALID_PIN'
+        assert PublicUserSignInAttempt.objects.filter(public_user=self.public_user).exists()
+
+    def test_verify_pin_wrong_pin_persists_attempt_increment(self, graphql_client_query):
+        # The verify call increments the attempts counter via verify(). The
+        # mutation wraps verify in transaction.atomic(); raising INVALID_PIN
+        # from inside the block previously rolled the increment back, so the
+        # PIN_MAX_ATTEMPTS guard was never reached.
```
tituomin: No need to explain bugs which were never live.

actions/tests/test_graphql_pledge.py
```diff
@@ -1099,13 +1108,760 @@ def test_set_user_data_with_bearer_token_authenticates(self, graphql_client_quer
...
+            variables={'email': 'alice@example.com', 'pin': raw_pin},
+        )
+
+        assert 'errors' in response
+        assert 'expired' in response['errors'][0]['message'].lower()
+        assert response['errors'][0]['extensions']['code'] == 'PIN_EXPIRED'
+
+    def test_verify_pin_last_allowed_attempt_succeeds(self, graphql_client_query, graphql_client_query_data):
+        # The Nth submission (where N == PIN_MAX_ATTEMPTS) must still be
+        # usable. A previous bug rejected it because verify() incremented
```
tituomin: No need to explain bugs which were never live.

actions/public_user_auth.py
Bhavatu: Not sure. More often if a user logs in with a anonymous user, they will have given the zipcode on commit, then they are asked if they want to log in/sign up

tituomin: Yes, so maybe worth checking where in the UI the potential verified_user.user_data(.zip_code) might come from, or maybe it's just copied from the existing data anyway...

actions/models/pledge.py
tituomin: This comment is focused on a line but I meant all the changes to this file related to signing in

aplans/graphene_views.py
```diff
@@ -212,13 +213,31 @@ def __init__(self, *args, **kwargs):
             kwargs['middleware'] = middleware
         super().__init__(*args, **kwargs)
 
+    # PublicUser auth-flow operations and bearer-varying reads must never be served
+    # from cache. Otherwise PIN emails would not be re-sent, freshly-issued
+    # userTokens would be replayed, and bearer-gated data could leak across callers.
+    PUBLIC_USER_AUTH_OPERATIONS = (
```
juyrjola: A better way is to ensure mutations are never cached

tituomin (approved):

aplans/schema_context.py
```diff
@@ -324,13 +324,8 @@ def get_cache_key_parts(self) -> list[str] | None:
...
-            self.set_reason('public-user bearer auth present')
-            return None
-
-        if 'publicUser' in (self.execution_context.query or ''):
-            self.set_reason('publicUser query is identity-sensitive and not cached')
+        if headers.get(PUBLIC_USER_TOKEN_HEADER):
+            self.set_reason('public-user token present')
             return None
 
         parts = [str(plan.identifier), plan.cache_invalidated_at.isoformat()]
```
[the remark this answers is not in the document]
Bhavatu: test_public_user_query_response_is_not_cached

src/aplans/settings.py
```diff
@@ -911,6 +912,7 @@ def parse_redirect_hostnames(redirect_hostnames: list[str]) -> tuple[tuple[str,
 # Reverse proxy stuff
 USE_X_FORWARDED_HOST = True
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+RATELIMIT_IP_META_KEY = 'aplans.ratelimit.client_ip_for_ratelimit'
```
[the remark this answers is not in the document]
Bhavatu: django-ratelimit 4.1.0 accepts `RATELIMIT_IP_META_KEY` in four shapes: a
callable, a dotted import path (any string containing `.`), a literal
`request.META` key, or unset (falls back to `REMOTE_ADDR`).

`'aplans.ratelimit.client_ip_for_ratelimit'` contains dots, so `_get_ip`
takes the dotted-path branch, imports it via `django.utils.module_loading.import_string`,
and calls the resolver with the request:

https://github.com/jsocol/django-ratelimit/blob/v4.1.0/django_ratelimit/core.py#L42-L44

Verified:

- `_get_ip(request)` with `X-Forwarded-For: 203.0.113.7, REMOTE_ADDR: 10.0.0.1`
  returns `203.0.113.7` (no `ImproperlyConfigured` raised).
- The signup / signIn / verifyPin test suites (`test_pledge.py`,
  `test_graphql_pledge.py`) all pass. Every one of those tests goes through
  `enforce_rate_limit()`; if the setting raised on each call they would all
  500.

No change needed.

src/admin_site/migrations/0020_backfill_primary_clientplan.py
```diff
@@ -0,0 +1,28 @@
...
+from django.db import migrations
+from django.db.models import Count
+
+
+def backfill_primary_clientplan(apps, schema_editor):
+    ClientPlan = apps.get_model('admin_site', 'ClientPlan')
+    Plan = apps.get_model('actions', 'Plan')
+
+    plans_with_one_client = Plan.objects.annotate(_n=Count('clients')).filter(_n=1).values_list('id', flat=True)
+    ClientPlan.objects.filter(plan_id__in=list(plans_with_one_client), is_primary=False).update(is_primary=True)
```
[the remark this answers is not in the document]
Bhavatu: We will need to assign them manually in the admin ui, and turning on the feature is not possible if primary client is not assigned

PR #632, opened 2026-07-03
Add single-tenant plan export command

bbliem:
Add an `export_plan` management command that exports one plan's data as dumpdata-format JSON for handing over to a tenant. Unlike the destructive `destructively_trim_db` + dumpdata approach (a denylist that must chase orphaned data to avoid leaking other tenants), this is an allowlist: it reuses the copying app's declarative clone structures to collect only plan-owned objects, so it cannot leak another tenant's rows by construction.

- Extract `iter_plan_owned_instances` from the copy unique-field validation helper so both copy and export share the plan-ownership traversal.
- `copying/export.py`: derive EXPORT_PLAN_STRUCTURE from PLAN_CLONE_STRUCTURE (re-including reports/links/change-logs that copy drops only to avoid reference remapping; PII gated behind flags); collect the Wagtail site, collection, page (MTI base + concrete rows) and media rows the clone structures skip; serialize with natural foreign keys.
- Close over reference-consistency gaps so the dump has no dangling FKs: indicators referenced by exported rows but not in plan.indicators (and the dimensions their values need), and images/documents referenced by pages or rich text (via FK and Wagtail ReferenceIndex) that live outside the plan's collections. Indicators/dimensions owned by another plan are skipped to preserve tenant isolation.
- --include-referenced-shared-objects opt-in does a one-level pull of referenced shared objects (orgs, persons, ...), following Organization ancestors transitively so the org tree isn't truncated.
- Add coverage, tenant-isolation, closure and round-trip tests, plus an architecture doc.

------

### Type of Change

### Testing

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -18,6 +18,7 @@ omit =
 show_missing = True
 exclude_lines =
     pragma: no cover
+    if TYPE_CHECKING:
     def __repr__
     if self.debug:
     if __name__ == .__main__.:
--- a/copying/export.py
+++ b/copying/export.py
@@ -1,0 +1,454 @@
+"""
+Single-tenant data export.
+
+Produce a self-contained JSON dump of one plan's data (and nothing belonging to any other
+tenant), in Django's standard serialization format -- the same format ``dumpdata`` emits and
+``loaddata`` consumes.
+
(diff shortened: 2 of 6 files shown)


tituomin (approved):
I rebased and it worked. Nice!

PR #634, opened 2026-07-03
Clean up orphaned data in destructively_trim_db for single-tenant exports

bbliem:
Trimming the DB to a single plan and running dumpdata left behind data that is not scoped to the retained plan, leaking other clients' content into the export. Add three cleanups to the trim command:

- delete_orphaned_translation_data() (always): sweep wagtail_localize rows whose translated object no longer exists. The trim deletes inside mute_signals(post_delete), which suppresses wagtail_localize's own cleanup_translation_on_delete handler, so this mirrors it after the fact.
- delete_orphaned_plan_pages() (always): delete PlanRootPage trees (any locale) and their Sites not belonging to a retained plan. Plan.delete() uses a bulk PageQuerySet.delete() that leaves a deleted plan's translated locale-tree pages (and Site) behind as live orphans.
- prune_unused_common_indicators() (behind --prune-shared-reference-data): delete common indicators not linked to a retained plan or a surviving indicator, then empty frameworks.

Extract the deletion-summary printing into helpers to keep handle() under the complexity limits.

While this removes some data that should not be in single-tenant exports, it is not sufficient. When I tested this with a churned client's data, there were still several records included that belonged to some other tenant for one reason or another. For cases like this, where we hand off data to customers, we need to be extra sure that we do not include data from other customers. A destructively_trim_db -> dumpdata approach seems to risky to me. Therefore, for data hand-off cases like this, I decided to go with a constructive approach that reuses some code from the copying app in order to only export data that is in fact reachable from the tenant's plan. There is a different PR for this: https://github.com/kausaltech/kausal-watch/pull/632

------

### Type of Change

### Testing

### DIFF
--- a/actions/management/commands/destructively_trim_db.py
+++ b/actions/management/commands/destructively_trim_db.py
@@ -28,6 +28,8 @@ from request_log.models import LoggedRequest
 from users.models import User
 
 if TYPE_CHECKING:
+    from uuid import UUID
+
     from django.db.models import Field, Model
 
     from orgs.models import OrganizationQuerySet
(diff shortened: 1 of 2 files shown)


Bhavatu:
Some merge conflicts to be resolved

